Issue #1, opened 2026-01-13
Build System: Unify CMakeLists.txt for dual-game build


spencerduncan:
## Goal
Modify the build system to support building both OoT and MM from the same CMakeLists.txt structure.

## Tasks
- [ ] Create root CMakeLists.txt that can build both games
- [ ] Create `games/oot/CMakeLists.txt` for OoT-specific build
- [ ] Create `games/mm/CMakeLists.txt` for MM-specific build  
- [ ] Unify cvar files (`CMake/lus-cvars.cmake` shared, game-specific cvars separate)
- [ ] Add CMake option to select which game(s) to build

## Acceptance Criteria
- `cmake -DGAME=oot` builds OoT executable
- `cmake -DGAME=mm` builds MM executable
- `cmake -DGAME=both` builds combined executable (future)

## References
- SoH CMakeLists.txt: `CMakeLists.txt`
- 2S2H CMakeLists.txt: `games/mm/CMakeLists.txt` (from 2ship2harkinian)

PR #124, opened 2026-01-23
Use Docker container for Linux builds instead of manual dependency installation

spencerduncan:
## Summary
Refactored the Linux build workflow to use a pre-built Docker container image instead of manually installing dependencies during each build. This simplifies the workflow, improves build consistency, and reduces build time by eliminating repetitive dependency compilation.

## Key Changes
- Changed runner from `ubuntu-22.04` to `ubuntu-latest` with a custom container image (`ghcr.io/spencerduncan/redshipblueship-build:latest`)
- Removed manual dependency installation step that used `apt-get` to install packages from `apt-deps.txt`
- Removed all custom dependency build steps:
  - SDL 2.30.3 compilation and installation
  - SDL_net 2.2.0 compilation and installation
  - tinyxml2 10.0.0 compilation and installation
  - libzip 1.10.1 compilation and installation
- Removed dependency caching logic (restore and save cache steps) as dependencies are now baked into the container image
- Kept ccache configuration for C/C++ compilation caching

## Benefits
- **Faster builds**: Dependencies are pre-compiled in the container, eliminating build time for SDL, SDL_net, tinyxml2, and libzip
- **Consistency**: All builds use the exact same dependency versions and configurations
- **Maintainability**: Dependency management is centralized in the container image rather than scattered across workflow steps
- **Cleaner workflow**: Reduces workflow file complexity and makes the build process more readable

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6524862995.zip)


spencerduncan (commented):
## Review: Migrate Linux CI to Docker Container

Overall this is a solid direction -- replacing ~80 lines of manual dependency installation with a pre-built container is a clear win for CI reliability and speed. However, there are several issues to address before merging.

---

### Branch needs rebase

The PR branch diverged from main at `d163e3dc` and is missing **18 commits** that have since landed on main, including:

- `36d8b103` -- Add `fetch-tags: true` to GitHub Actions checkout steps
- `eba14a19` -- Sync Dockerfile apt packages with `apt-deps.txt`
- `0c07e392` -- Add smoke test to Docker image build
- `1ff03269` -- Add optional lld-link support for faster Windows builds (adds `llvm` to Windows choco install)

The consequence visible in the diff: the `build-linux` checkout step is **missing `fetch-tags: true`** compared to main. This will cause version detection to break if it relies on git tags. Please rebase onto main to incorporate these commits and resolve the conflicts they'll surface (especially in the Dockerfile).

---

### Issue: `latest` tag is non-deterministic (medium)

**`.github/workflows/generate-builds.yml`**, build-linux container reference:
```yaml
container:
  image: ghcr.io/spencerduncan/redshipblueship-build:latest
```

Using `:latest` means any push to the container registry silently changes what CI builds against. Two runs of the same commit can produce different results. Consider either:
1. Pinning to a digest (`@sha256:...`)
2. Using a versioned tag (e.g., `:v1`, `:2024.03`) that is updated deliberately
3. At minimum, tagging with the git SHA that built the image

This is the single biggest reproducibility concern in this PR.

---

### Issue: Dockerfile diverges from `apt-deps.txt` (medium)

The PR branch Dockerfile inlines all packages directly (lines 37-55) instead of using the `COPY apt-deps.txt` + `tr/xargs` pattern that was introduced on main by `eba14a19`. The `lsb-release` package is now listed in both `apt-deps.txt` (line 1) and inline in the Dockerfile (line 59), and future edits to `apt-deps.txt` will not be reflected in the Docker image. This is a regression -- the single-source-of-truth pattern should be preserved. This will likely be fixed naturally by rebasing onto main.

---

### Issue: `gamecontrollerdb.txt` URL inconsistency (low, pre-existing)

**`games/mm/CMakeLists.txt:819`** (on main) uses `gabomdq/SDL_GameControllerDB` while `games/oot/CMakeLists.txt:822` (on main) uses `mdqinc/SDL_GameControllerDB`. The PR branch fixes both to use `mdqinc` (the current canonical repo name -- `gabomdq` redirects there). This is a good cleanup. The Docker image also uses `mdqinc`. Consistent.

---

### sohIcon.png fallback -- looks correct (positive)

**`CMakeLists.txt:371-388`**: The ImageMagick + file-copy fallback is well-structured:
1. Try ImageMagick `convert` to resize the icon to 512x512
2. Check `RESULT_VARIABLE` for failure (previously ignored)
3. If the output file does not exist (ImageMagick missing or failed), copy the source icon as-is

One minor note: the fallback copies the original icon without resizing. If a 512x512 icon is strictly required for AppImage packaging, this could be an issue. If the original is already 512x512 or if the AppImage tooling is tolerant of other sizes, this is fine.

---

### ctest label change -- correct but worth documenting (positive)

**`CMake/SingleExecutable.cmake:195`**: Renaming `LABELS "redship"` to `LABELS "integration"` and using `--label-exclude "^integration$"` instead of `--label-regex "^redship$"` in the workflow inverts the test filter logic.

- **Before**: run only tests labeled `redship` (which was all integration tests)
- **After**: run all tests *except* those labeled `integration`

This means if someone adds a new test without a label, it will now run in CI automatically (previously it would be silently skipped). This is arguably the better default, but it's a semantic change worth noting in the PR description.

---

### Supply chain / security observations

1. **Container image trust**: `ghcr.io/spencerduncan/redshipblueship-build` is owned by the same GitHub user/org as the repo. The Dockerfile source is in-tree at `.github/docker/build-base.Dockerfile`. The supply chain is transparent -- good.

2. **No `credentials` or `options` on the container**: The `container:` block doesn't specify `credentials` for pulling, which means the image must be public on GHCR. Confirm this is the case, otherwise PR builds from forks will fail.

3. **Ubuntu base image**: `ubuntu:22.04` is pulled from Docker Hub without digest pinning. Standard practice but worth noting for auditors.

---

### Dependencies covered by Docker -- verified

All removed workflow steps are accounted for in the Dockerfile:

| Removed workflow step | Dockerfile equivalent |
|---|---|
| `apt-get install $(cat apt-deps.txt)` | Lines 35-51 (inline) + `apt-deps.txt` packages |
| Install SDL2 2.30.3 | Lines 56-64 (same version, same `--enable-hidapi-libusb` flag) |
| Install SDL2_net 2.2.0 | Lines 67-75 (same version) |
| Install tinyxml2 10.0.0 | Lines 78-87 (same version, same `-DCMAKE_POSITION_INDEPENDENT_CODE=ON`) |
| Install libzip 1.10.1 | Lines 89-103 (same version, same crypto-disabled flags) |
| Deps cache save/restore | Eliminated (deps are baked into image) |

---

### Removed cache steps -- clean

The `Restore Cached deps folder` and `Save Cache deps folder` steps are correctly removed. The Docker image replaces this caching strategy entirely. The ccache action is preserved, which is correct (ccache caches compilation output, not dependencies).

---

### OTR generation -- unchanged

The `generate-rsbs-otr` job is not modified and still runs on `ubuntu-22.04` with manual dependency installation. This is fine -- it could be migrated to Docker separately in a follow-up.

---

### Windows build -- unchanged (confirmed)

The `build-windows` job has no changes in this diff. Verified.

---

### Summary of action items

1. **Must fix**: Rebase onto main to pick up `fetch-tags`, Dockerfile sync, smoke test, and other commits
2. **Should fix**: Pin the container image to a digest or versioned tag instead of `:latest`
3. **Nice to have**: Update PR description to note the ctest label semantic change

spencerduncan (commented):
## Triage Review — 2026-04-19

Reviewing after ~4 weeks of inactivity since the last commit (`fe8ce4408`, 2026-03-24) and the prior owner review. Bottom line: **the direction is still correct and still needed** (main's `build-linux` is unchanged from what this PR replaces), but the PR has a latent bug that would have failed any rebase attempt and needs a design fix, not just a nudge.

### Verdict: REQUEST CHANGES — rebase + fix bootstrap ordering, don't close

Only **7 commits behind main now** (not as stale as it looks). Rebase cost is low. But the `build-linux` failure is a real design problem that will resurface even after rebase.

---

### The `build-linux` icon failure is a design flaw, not a CI flake

**Root cause (verified):**
- `games/oot/macosx/sohIcon.png` is **1024×1024** (confirmed via `file`)
- [CMakeLists.txt:394-407](CMakeLists.txt:394) resizes with ImageMagick to 512×512 when `ImageMagick_FOUND`, then falls back to `file(COPY ...)` of the **original 1024×1024** source if ImageMagick is missing or fails
- linuxdeploy ([CMake/Packaging.cmake:55-84](CMake/Packaging.cmake:55)) only accepts standard AppImage sizes (512 being the canonical large one); 1024×1024 is rejected

**Why it's firing on this PR specifically:** chicken-and-egg with the container image.

- [build-docker-image.yml:82,90](.github/workflows/build-docker-image.yml:82) only pushes + tags `:latest` on pushes to `main` — PR builds *load* the image but never publish it
- `generate-builds.yml` in this PR pulls `ghcr.io/spencerduncan/redshipblueship-build:latest`, which is whatever was last built from `main` — **without the `imagemagick` line this PR adds**
- So inside the container, `ImageMagick_FOUND` is false → fallback fires → 1024×1024 icon → linuxdeploy rejects it

Two possible fixes, in order of preference:

1. **Commit a pre-resized 512×512 icon in the repo** and drop the ImageMagick dependency entirely. Removes a build-time dep, removes the fallback, removes the bootstrap problem. This is roughly 3 lines of change in CMakeLists.txt after checking in `games/oot/macosx/sohIcon-512.png`.
2. **Make the fallback also resize** (or make it fatal so failures are loud instead of producing broken artifacts). Still leaves the container-bootstrap problem — first merge to main will need a manual docker image rebuild before this PR can land.

### Prior review items still unaddressed

Reiterating from spencerduncan's 2026-03-24 review — none of these were fixed:

1. **`:latest` tag is non-deterministic** — pin to a digest (`@sha256:...`) or SHA tag. Has the pleasant side effect of solving the bootstrap problem above: a Renovate-style tag bump flow would let the PR update both the Dockerfile *and* the tag in the same change, tested together.
2. **Dockerfile no longer uses `apt-deps.txt` as single source of truth** — likely resolves naturally on rebase, but needs verification.
3. **ctest semantics inverted** (`--label-regex "^redship$"` → `--label-exclude "^integration$"`) — worth a line in the PR description; new unlabeled tests will now run by default.

### What's still OK

- Removed deps (SDL2, SDL2_net, tinyxml2, libzip) are all accounted for in [build-base.Dockerfile](.github/docker/build-base.Dockerfile) at matching versions and flags
- ccache action correctly preserved (ccache caches compilation, not deps)
- `gamecontrollerdb.txt` URL normalized to `mdqinc/` across OoT and MM (good cleanup)
- `generate-rsbs-otr` intentionally unchanged — can migrate in a follow-up

### Recommended path

1. Rebase onto `main` (7 commits, low churn)
2. Replace the ImageMagick dance with a committed 512×512 icon (fix #1 above)
3. Pin the container image reference
4. Push — expect the first run to still fail because `:latest` on GHCR predates the `imagemagick` fix, but with a pinned image from a PR-built SHA tag, you can test the full chain end-to-end

If rebasing + those fixes takes more than an hour, close and restart — but the code here is salvageable and the alternative is rewriting the same thing.

spencerduncan:
Addressed the REQUEST CHANGES feedback from the 2026-04-19 triage review.

**Rebased** onto `origin/main` cleanly (no conflicts); branch is now current.

**Fixed the `build-linux` icon failure** per Option 1 from the review — checked in a pre-resized 512×512 `games/oot/macosx/sohIcon-512.png` and replaced the ImageMagick `execute_process` + file-copy fallback in `CMakeLists.txt` with a single `configure_file(... COPYONLY)`. Also dropped `imagemagick` from `.github/docker/build-base.Dockerfile`.

This resolves the chicken-and-egg bootstrap problem: the container no longer needs ImageMagick, so PR builds don't depend on a `:latest` image built from a future `main` merge. The original 1024×1024 `sohIcon.png` is kept intact for the macOS `CreateOSXIcons` target, which needs it at `@2x` resolution.

**Still open** (not addressed in this push — happy to tackle in follow-ups or this PR if preferred):
- Pinning the container image reference away from `:latest` to a digest or SHA tag
- Noting the ctest `--label-regex` → `--label-exclude` semantic change in the PR description

All five original commits preserved; the fix is a single new commit on top (`2e18934851`).

spencerduncan (commented):

# PR #124 Review — Use Docker container for Linux builds

**Verdict: REQUEST CHANGES**

Repo: `spencerduncan/redshipblueship`
Branch: `claude/migrate-linux-ci-docker-ZYCaJ` @ `2e18934`
Review date: 2026-04-24

The April 20 rebase landed cleanly and the pre-sized 512×512 icon fix is a
genuine improvement over the ImageMagick-at-configure-time dance. But the
PR ships a silently-broken CI step that I can't let through, plus two items
the author already flagged as open.

## Blocking

### 1. ctest label filter is broken — integration tests are now silently skipped

`CMake/SingleExecutable.cmake` labels the integration suite
`BootOoT BootMM SwitchOoTMM SwitchMMOoT Roundtrip Context AllTests` as
`LABELS "integration"`, but `.github/workflows/generate-builds.yml` still
invokes:

```
ctest --test-dir build-cmake --output-on-failure --label-regex "^redship$"
```

Nothing in the suite matches `^redship$` anymore, so ctest will run zero
tests and exit 0. Green build, no test coverage. The "Run tests" step
becomes a liar.

Commit `1c32bfe` ("Fix ctest to use --label-regex after rebase") reverted
the workflow side of the earlier `--label-exclude` change but missed that
the CMake label itself had already been renamed to `integration`. Pick one:

- Rename the label back: `LABELS "redship"` in SingleExecutable.cmake, or
- Update the workflow: `--label-regex "^integration$"`, or (preferred)
- Go back to `--label-exclude "^integration$"` and keep the inverted
  semantics — but then also update the PR description per the author's own
  2026-03-24 note.

Either way, verify locally with `ctest -N --label-regex ...` before pushing
that the expected tests actually list.

### 2. Container image still pinned to `:latest`

`generate-builds.yml` build-linux:
```
container:
  image: ghcr.io/spencerduncan/redshipblueship-build:latest
```

The author flagged this on 2026-03-24 and again on 2026-04-20 and explicitly
left it for follow-up. I don't think it should merge in that state — this
is the whole reproducibility story of the PR. A push to `main` that
rebuilds the base image silently changes what every PR CI run executes
against. That's the failure mode Docker containers are supposed to prevent.

Minimum acceptable: pin to a `:sha-<commit>` tag emitted by
`build-docker-image.yml` and bump it in the same PR as Dockerfile changes.
Better: pin to `@sha256:...` digest. The image is public GHCR under the
same owner, so auth is not a blocker.

While you're there, `FROM ubuntu:22.04` in `build-base.Dockerfile` is also
unpinned — lower severity since the base rarely moves in breaking ways,
but worth considering for an audit trail.

## Non-blocking notes

**Runner change, `ubuntu-22.04` → `ubuntu-latest`**: Not mentioned in the
PR description. Since the build runs inside the container, the runner OS
mostly affects the Docker daemon and the checkout host, so this is
low-risk. Still worth a one-line note so future-you remembers why.

**No Docker layer cache in the workflow**: The `container:` directive
pulls the whole image once per job; GHA's `jobs.<id>.container` doesn't
use `actions/cache` for layers, and adding `docker/build-push-action`
caching only helps `build-docker-image.yml`, not this workflow. Since the
consumer here just pulls a prebuilt image, this is fine — GHCR pulls are
fast and the image is ~700MB-ish of deps. Not a blocker; noting so it
doesn't surprise anyone.

**Pull failure behavior**: GitHub Actions fails the job fast if the
`container:` image can't be pulled, with a clear error. No hang risk. OK.

**Apt-deps parity (Dockerfile ↔ apt-deps.txt)**: Verified on the branch.
`build-base.Dockerfile` uses `COPY .github/workflows/apt-deps.txt` and
installs via `tr ' ' '\n' < /tmp/apt-deps.txt | xargs apt-get install -y`.
Single source of truth is preserved — this was a concern in the March
review and is now resolved. The Dockerfile also inlines `build-essential`,
`cmake`, `git`, `zlib1g-dev`, `libbz2-dev`, `liblzma-dev`, `libzstd-dev`,
`ccache`, `curl`, which are intentionally not in apt-deps.txt (commented
as such). Fine.

**Icon fix**: `configure_file(... sohIcon-512.png ... COPYONLY)` + committed
`games/oot/macosx/sohIcon-512.png` + `.gitignore` exception is the right
shape. Removes ImageMagick, removes the silent-corruption fallback path,
and sidesteps the bootstrap chicken-and-egg. Good.

**gamecontrollerdb.txt**: The CMake logic in both `games/oot/CMakeLists.txt`
and `games/mm/CMakeLists.txt` correctly prefers the pre-cached
`/opt/redshipblueship/gamecontrollerdb.txt` and falls back to curl.
URL standardized to `mdqinc/SDL_GameControllerDB` across both. Clean.

**Rebase freshness**: 4 days old (Apr 20). Author confirmed clean rebase
onto `origin/main`. Assume current unless checks say otherwise.

## Files changed (7)

- `.github/workflows/generate-builds.yml` — build-linux job simplified (+3/-87)
- `CMake/SingleExecutable.cmake` — label rename `redship` → `integration` ⚠️
- `CMakeLists.txt` — ImageMagick dance → configure_file
- `games/mm/CMakeLists.txt` — gamecontrollerdb.txt pre-cache lookup
- `games/oot/CMakeLists.txt` — gamecontrollerdb.txt pre-cache lookup
- `games/oot/.gitignore` — allow `sohIcon-512.png`
- `games/oot/macosx/sohIcon-512.png` — new (64KB, pre-sized)

## Summary of required changes before merge

1. **Fix the ctest label/filter mismatch.** Currently the Linux CI runs
   zero integration tests. Blocking.
2. **Pin the container image reference away from `:latest`.** SHA tag or
   digest. Blocking for the PR's own stated reproducibility goal.
3. (Nice to have) Note the runner change and the label semantic change in
   the PR description.

spencerduncan:
Closing as superseded. The Docker Linux CI effort landed piecemeal in other PRs: #196 synced Dockerfile packages with apt-deps.txt, #222 hardened the Docker worker scripts, #193 added smoke tests to the Docker build workflow. The original consolidated approach in this PR is no longer needed.

PR #190, opened 2026-02-01
Add integration test framework with GameInteractor hooks

spencerduncan:
## Summary
This PR introduces a comprehensive integration test framework that allows testing actual game boot sequences using GameInteractor hooks. Integration tests differ from unit tests by actually booting the games and verifying they reach expected states (title screen, file select, etc.) rather than just testing infrastructure.

## Key Changes

- **New Integration Test Infrastructure** (`integration_test_hooks.h/cpp`)
  - State management for integration tests with atomic flags for thread-safe boot detection
  - Support for multiple test modes (boot OoT, boot MM, game switching)
  - Timeout management and exit signaling

- **GameInteractor Hook Registration**
  - OoT: Registers hooks for `OnZTitleInit` and `OnPresentFileSelect` to detect boot completion
  - MM: Registers hooks for `OnConsoleLogoUpdate` and `OnGameStateMainStart` to detect boot completion
  - Hooks are registered in each game's `Game_Init` function when integration test mode is active

- **Test Runner Enhancements** (`test_runner.cpp/h`)
  - New `TestRunner_ParseIntegrationArgs()` to parse `--integration-test` flag
  - `TestRunner_SetupIntegrationTest()` to initialize integration test mode
  - Integration test descriptors separate from unit tests
  - Result reporting via `TestRunner_GetIntegrationTestResult()`

- **Main Loop Integration** (`rsbs/src/main.cpp`)
  - Detects `--integration-test` flag and sets up test mode before game initialization
  - Overrides game selection based on integration test requirements
  - Filters integration test flags from game argv
  - Returns integration test result as exit code

- **CMake Configuration** (`CMake/SingleExecutable.cmake`)
  - Adds integration test targets with 120-second timeout
  - Separates integration tests from unit tests with distinct labels
  - Includes GameInteractor headers for hook registration

- **CI/CD Pipeline** (`.github/workflows/integration-tests.yml`)
  - New GitHub Actions workflow for integration tests on Linux
  - Builds dependencies (SDL2, SDL_net, tinyxml2, libzip)
  - Runs with Xvfb for headless rendering
  - Includes timeout protection (120 seconds per test)
  - Caches OTR generation and dependencies for faster runs

## Implementation Details

- Integration tests use `Combo_RequestGameSwitch()` to cleanly exit the game after boot detection
- Hooks use frame counting to ensure stable boot before signaling completion
- Atomic flags ensure thread-safe communication between game threads and test framework
- Tests are marked with "integration" label in CMake for easy filtering
- Separate from unit tests which use `--test` flag; integration tests use `--integration-test` flag

https://claude.ai/code/session_01R1CM7nNiYRQFqQwbbwVAA3

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6074227236.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6073957119.zip)

### DIFF
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -301,7 +301,7 @@ jobs:
         CC: gcc-11
         CXX: g++-11
     - name: Run tests
-      run: ctest --test-dir build-cmake --output-on-failure --label-exclude "^integration$"
+      run: ctest --test-dir build-cmake --output-on-failure --label-regex "^redship$"
     - name: Package
       run: |
         (cd build-cmake && cpack -G External)
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -301,7 +301,7 @@ jobs:
         CC: gcc-11
         CXX: g++-11
     - name: Run tests
-      run: ctest --test-dir build-cmake --output-on-failure --label-regex "^redship$"
+      run: ctest --test-dir build-cmake --output-on-failure --label-exclude "^integration$"
     - name: Package
       run: |
         (cd build-cmake && cpack -G External)
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,272 @@
+name: integration-tests
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'LICENSE*'
+      - '.gitignore'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'LICENSE*'
+      - '.gitignore'
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  integration-tests-linux:
+    needs: generate-rsbs-otr
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Git Checkout
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+        fetch-depth: 1
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y $(cat .github/workflows/apt-deps.txt)
+        # Xvfb and mesa for headless rendering
+        sudo apt-get install -y xvfb libgl1-mesa-dri libgl1-mesa-glx mesa-utils
+
+    - name: Configure ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        save: ${{ github.ref_name == github.event.repository.default_branch }}
+        key: ${{ runner.os }}-integration-ccache-${{ github.ref }}
+        restore-keys: |
+          ${{ runner.os }}-integration-ccache-${{ github.ref }}
+          ${{ runner.os }}-integration-ccache-refs/heads/main
+          ${{ runner.os }}-integration-ccache
+
+    - name: Restore Cached deps folder
+      id: restore-cache-deps
+      uses: actions/cache/restore@v4
+      with:
+        key: ${{ runner.os }}-deps-v1-${{ hashFiles('.github/workflows/apt-deps.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-deps-v1-
+        path: deps
+
+    - name: Create deps folder
+      run: mkdir -p deps
+
+    - name: Install latest SDL
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        if [ ! -d "deps/SDL2-2.30.3" ]; then
+          wget https://github.com/libsdl-org/SDL/releases/download/release-2.30.3/SDL2-2.30.3.tar.gz
+          tar -xzf SDL2-2.30.3.tar.gz -C deps
+        fi
+        cd deps/SDL2-2.30.3
+        if [ ! -f "Makefile" ]; then
+          ./configure --enable-hidapi-libusb
+        fi
+        make -j 10
+        sudo make install
+        sudo cp -av /usr/local/lib/libSDL* /lib/x86_64-linux-gnu/
+
+    - name: Install latest SDL_net
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        if [ ! -d "deps/SDL2_net-2.2.0" ]; then
+          wget https://www.libsdl.org/projects/SDL_net/release/SDL2_net-2.2.0.tar.gz
+          tar -xzf SDL2_net-2.2.0.tar.gz -C deps
+        fi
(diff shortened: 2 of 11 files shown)


spencerduncan (commented):
## Code Review: PR #190 — Add integration test framework with GameInteractor hooks

Overall this is a well-structured PR that cleanly separates integration tests (which boot the actual game) from the existing unit tests. The architecture of having each game register its own hooks in `GameExports_SingleExe.cpp` while keeping shared state in `integration_test_hooks.cpp` is sound. That said, there is one blocking issue with the exit mechanism, and several smaller items worth addressing.

---

### Blocking

**1. Integration test exit mechanism does not work correctly**

`integration_test_hooks.cpp:80` — When boot is detected, `IntegrationTest_SignalBootComplete()` calls `Combo_RequestGameSwitch()` to break out of the game's run loop. However, after `ops->run()` returns in `rsbs/src/main.cpp:335`, the main loop at lines 351-353 hits:

```cpp
} else if (Combo_IsGameSwitchRequested()) {
    nextGame = Game_GetOther(GameRunner_GetActive(&runner));
}
```

This switches to the *other* game instead of exiting the `while (keepRunning)` loop. The test would boot the other game and likely hang until the CI `timeout` kills it, resulting in a spurious failure (exit code 124) rather than a clean pass.

`IntegrationTest_ExitRequested()` is defined in `integration_test_hooks.cpp:94` but never checked in the main loop. The fix should add an early check in the main loop:

```cpp
// After ops->run() returns, before the switch logic:
if (TestRunner_IsIntegrationTestMode() && IntegrationTest_ExitRequested()) {
    keepRunning = false;
    continue;
}
```

Or alternatively, don't use `Combo_RequestGameSwitch()` at all — set `keepRunning` via a different mechanism (e.g., a global flag checked in the main loop, or call `exit(0)` directly from the hook after recording the result).

---

### Non-blocking

**2. `IntegrationTest_RegisterHooks()` in `main.cpp` is a no-op — potentially confusing**

`rsbs/src/main.cpp:355-358` — After the MM switch, `IntegrationTest_RegisterHooks(GAME_MM)` is called. But this function (`integration_test_hooks.cpp:97-113`) only prints log messages and doesn't actually register any hooks. The actual registration already happened inside `MM_Game_Init()` which was called by `GameRunner_SwitchTo()`. This is fine functionally but the call site reads as if it's doing something important. Consider either removing the call or renaming the function to `IntegrationTest_LogHookStatus()` to clarify intent.

**3. Duplicate `#include <ship/Context.h>` in MM GameExports**

`games/mm/2s2h/GameExports_SingleExe.cpp:19,24` — `<ship/Context.h>` is included twice. This was pre-existing (also present on `main`), but since you're editing the include block anyway, it would be good to clean up.

**4. `static int frameCount` in lambda captures will accumulate across game restarts**

`games/mm/2s2h/GameExports_SingleExe.cpp:102,116` — The `static int frameCount` variables inside the lambdas persist for the lifetime of the process. If the game were ever re-initialized within the same process (e.g., for a switch test), these counters would not reset. For the current boot-only tests this is fine, but it would be a latent bug for `INT_TEST_SWITCH_*` modes. Consider using `std::atomic<int>` in the anonymous namespace instead, and resetting them in `MM_RegisterIntegrationTestHooks`.

**5. Multiple hooks can both signal boot completion**

`games/mm/2s2h/GameExports_SingleExe.cpp:100-124` and `games/oot/soh/GameExports_SingleExe.cpp:57-76` — Both OoT and MM register two hooks that each call `IntegrationTest_SignalBootComplete()`. `SignalBootComplete` calls `Combo_RequestGameSwitch()` on each invocation. If both hooks fire (which is likely — `OnConsoleLogoUpdate` fires before `OnGameStateMainStart`), `Combo_RequestGameSwitch()` is called twice. This appears safe since the flag is a simple bool, but it's worth noting. Adding a guard like `if (sBootPassed.load()) return;` at the top of `IntegrationTest_SignalBootComplete` would make this more robust.

**6. `kBootTimeoutMs` is defined but never used**

`integration_test_hooks.cpp:34` — `constexpr int kBootTimeoutMs = 60000` is defined, and `IntegrationTest_GetTimeoutMs()` returns it, but nothing calls `IntegrationTest_GetTimeoutMs()`. Timeouts are handled externally by `timeout 120` in the CI workflow and `TIMEOUT 120` in CMake. This is dead code — either use it or remove it.

**7. CI workflow: `generate-rsbs-otr` job defined after the job that depends on it**

`.github/workflows/integration-tests.yml` — The `generate-rsbs-otr` job is defined at line 206 after `integration-tests-linux` which declares `needs: generate-rsbs-otr` at line 24. YAML job ordering doesn't matter for GitHub Actions (it uses `needs` for dependencies), so this is purely a readability issue. Consider reordering so `generate-rsbs-otr` appears first.

**8. CI workflow: archive check loop doesn't `break` on first match**

`.github/workflows/integration-tests.yml` lines 149-161 — The archive check loops through multiple directories but doesn't break after finding a match. If `oot.o2r` exists in both `build-cmake` and `build-cmake/soh`, the loop continues unnecessarily. This is cosmetic/harmless but a `break` would be cleaner.

**9. Mixed `printf`/`fprintf(stderr)` for test output**

Integration test output uses `printf` (stdout) in `test_runner.cpp` and `integration_test_hooks.cpp`, but `fprintf(stderr)` in the game-specific hooks (`GameExports_SingleExe.cpp`). In CI, both streams are merged, but if someone redirects stdout, the test output would be split across streams. Consider standardizing on `stderr` for all integration test diagnostic output.

**10. CMake: integration tests registered unconditionally under `BUILD_TESTING`**

`CMake/SingleExecutable.cmake:212-213` — `IntBootOoT` and `IntBootMM` are registered as ctest targets unconditionally. They will always appear in `ctest --label-regex integration` but will always fail in environments without game archives. The label separation (`"integration"` vs `"redship"`) mitigates this since unit tests use `--label-regex "^redship$"`, but it's worth documenting that integration tests require external archives.

---

### Things that look correct

- The `#ifdef RSBS_SINGLE_EXECUTABLE` guards are present on both `GameExports_SingleExe.cpp` files (wrapping the entire file). The new hook code is inside these guards. Good.
- The MM boot order fix (boot OoT first for `Ship::Context`, then switch to MM) at `main.cpp:274-280` is correct and consistent with the existing architecture.
- The ctest labels (`"redship"` for unit tests, `"integration"` for integration tests) are correct. The CI unit test step uses `--label-regex "^redship$"` which correctly excludes integration tests.
- `--integration-test` arg parsing and stripping in the argv filter loop is correct.
- The `std::atomic` usage for `sTestMode`, `sBootPassed`, `sExitRequested`, and `sBootedGame` is appropriate for cross-thread signaling (hooks fire on the game thread, main loop runs on the main thread).
- OoT hooks (`OnZTitleInit`, `OnPresentFileSelect`) and MM hooks (`OnConsoleLogoUpdate`, `OnGameStateMainStart`) are verified to exist in the respective GameInteractor implementations.

PR #217, opened 2026-03-24
Unstub MM AudioLoad and bring real load.c into build (#157)

### LINKED ISSUE #157: MM Audio system init for cross-game switch
## Problem
When switching from OoT to MM, `MM_AudioMgr_Init` crashes at `AudioSfx_Init → AudioThread_ScheduleProcessCmds → osSendMesg` with SIGSEGV (NULL message queue pointer).

## Stack trace
```
1 osSendMesg (+0x4)
2 AudioThread_ScheduleProcessCmds (+0x3F)
3 AudioSfx_Init (+0x11)
4 MM_AudioMgr_Init (+0x9C)
5 MM_Game_Init (+0x248)
6 GameRunner_SwitchTo (+0x8F)
```

## Root cause
MM's audio subsystem assumes it owns the audio thread and message queues. In single-exe mode, OoT's audio system was using these and they're in an indeterminate state after OoT suspends.

## Dependencies
- Ship::Context must be alive (fixed by #153)
- `CVarGetFloat` calls in AudioMgr_Init need Ship::Context (now works)
- Audio thread message queues need to be properly initialized or reset

## Approach
- Trace `MM_AudioMgr_Init` → `AudioSfx_Init` → `AudioThread_ScheduleProcessCmds` to find which queue is NULL
- Either reinitialize the audio subsystem cleanly, or properly hand off from OoT's audio state
- Unit test: mock message queue init, verify AudioMgr_Init succeeds

## Context
Blocked by: #153 (lifecycle system)
Part of: #145 (single-exe architecture)

spencerduncan:
## Summary
- Remove `load.c` exclusion from `games/mm/CMakeLists.txt` 
- Remove 26 AudioLoad stub functions from `mm_stubs.c`
- Remove `func_8018FA60` stub (had wrong 3-param signature vs real 4-param — undefined behavior at `thread.c:224`)
- Remove conflicting global variable stubs

The real implementations in `games/mm/src/audio/lib/load.c` now compile into the single-exe build, giving MM actual audio loading.

Fixes #157

## Test plan
- [ ] Build compiles with no duplicate symbol errors
- [ ] 7/7 unit tests pass
- [ ] MM has audio after game switch (requires ROMs)

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6076117729.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6075333031.zip)


spencerduncan:
## Build Failure: Multiple Definition + Missing Symbol

The naive approach of just un-excluding `load.c` and removing stubs doesn't work because:

### 1. Symbol collisions (multiple definition)
MM's `load.c` has ~30 internal functions (`AudioLoad_InitSlowLoads`, `AudioLoad_GetFontSample`, `AudioLoad_RelocateSample`, etc.) with the **same names** as OoT's `audio_load.c`. The MM namespace tool only prefixes the `MM_AudioLoad_*` entry points, not these shared internal helpers.

### 2. Missing symbol
`MM_osCartRomInit` — referenced by `load.c` but not defined anywhere (was either in stubs or never implemented).

### Fix needed
The MM namespace tool needs to prefix ALL functions in `load.c`, not just the public API. This is a CMake/build-system change in how the MM namespace prefixing works for the audio subsystem. Alternatively, we could wrap the internal functions with `static` or namespace them manually.

This is a bigger task than the original stream scope. Keeping the PR open for a follow-up.

Issue #31, opened 2026-01-18
Cross-Game Context and Switching Infrastructure


spencerduncan:
## Context

With both games in a single executable (#30), we need infrastructure for switching between them seamlessly. This implements the OoTMM-style context exchange.

Adapts concepts from #14 (Frozen State Manager).

### ComboContext Structure
```cpp
// src/common/context.h
#pragma once

enum class Game { None, OoT, MM };

struct ComboContext {
    // Magic validation
    char magic[8] = "OoT+MM<3";
    uint32_t version = 1;

    // Switch control
    bool switchRequested = false;
    Game targetGame = Game::None;
    uint16_t targetEntrance = 0;

    // Source state (for return trips)
    Game sourceGame = Game::None;
    uint16_t sourceEntrance = 0;

    // Shared persistent state
    uint32_t sharedFlags[64];       // Cross-game event flags
    uint16_t sharedItems[32];       // Shared inventory items

    // Save slot tracking
    int32_t saveSlot = 0;
};

extern ComboContext gComboCtx;
extern Game gCurrentGame;
```

### Game Switch Function
```cpp
// src/common/switch.cpp
#include "context.h"
#include "game_api.h"

void ComboGameSwitch(Game target, uint16_t entrance) {
    // Freeze current game state
    if (gCurrentGame == Game::OoT) {
        OoT::FreezeState(&gComboCtx);
    } else if (gCurrentGame == Game::MM) {
        MM::FreezeState(&gComboCtx);
    }

    // Store source for return trip
    gComboCtx.sourceGame = gCurrentGame;
    gComboCtx.targetEntrance = entrance;

    // Signal switch
    gComboCtx.switchRequested = true;
    gComboCtx.targetGame = target;
}

void ProcessGameSwitch() {
    if (!gComboCtx.switchRequested) return;

    gCurrentGame = gComboCtx.targetGame;
    gComboCtx.switchRequested = false;

    // Resume target game
    if (gCurrentGame == Game::OoT) {
        OoT::ResumeFromContext(&gComboCtx);
    } else {
        MM::ResumeFromContext(&gComboCtx);
    }
}
```

### IGameAdapter Interface
```cpp
// Each game implements these for integration
class IGameAdapter {
    virtual void Init() = 0;
    virtual void Run() = 0;
    virtual void FreezeState(ComboContext* ctx) = 0;
    virtual void ResumeFromContext(ComboContext* ctx) = 0;
};
```

## Files to Create

[...]

PR #220, opened 2026-04-19
Add error handling policy documentation

spencerduncan:
## Summary
- Adds `docs/error-handling.md` establishing consistent error handling patterns across the codebase (C error codes, C++ exceptions/`std::optional`/`std::expected`, RAII, libultraship logging).
- Links the new doc from `README.md`.
- Blocks [#31](https://github.com/spencerduncan/redshipblueship/issues/31) (Cross-Game Context), which should be implemented against these patterns.

## Test plan
- [ ] Docs-only change — verify rendered markdown on GitHub reads correctly.
- [ ] Confirm README link resolves to `docs/error-handling.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,14 @@ redshipblueship/
 └── OTRExporter/      # OTR generation
 ```
 
+## Documentation
+
+- [Building](docs/BUILDING.md) - Build instructions
+- [Error Handling Policy](docs/error-handling.md) - Error handling patterns and guidelines
+- [Modding](docs/MODDING.md) - Modding guide
+- [Custom Music](docs/CUSTOM_MUSIC.md) - Custom music support
+- [Credits](docs/CREDITS.md) - Full credits
+
 ## Quick Start (Once Built)
 
 1. Place `oot.z64` and `mm.z64` ROMs in the game directory
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -1,0 +1,60 @@
+# Error Handling Policy
+
+## Principles
+
+1. **No silent failures** - Every error must be surfaced
+2. **Appropriate mechanisms** - Match error handling to context
+3. **Predictable behavior** - Same patterns across codebase
+
+## Patterns by Context
+
+### C APIs (Cross-DLL boundaries)
+- Return error codes (int or enum)
+- Use out parameters for results
+- Document error codes in header comments
+
+### C++ Internals
+- Use exceptions for programming errors (logic bugs)
+- Use `std::optional<T>` for expected "not found" cases
+- Use `std::expected<T, E>` (C++23) or custom `Result<T>` for recoverable errors
+
+### Resource Acquisition
+- Always use RAII (`ScopedLibrary`, smart pointers)
+- Cleanup in destructors, not manual calls
+- Document ownership transfers
+
+### Logging
+- Use libultraship logging, not `std::cerr`
+- Debug output must be gated by log level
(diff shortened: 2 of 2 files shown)


spencerduncan (commented):
## Review: Error Handling Policy Documentation

**Verdict: COMMENT** — the doc is useful and substantively accurate for this codebase, but there are a few inconsistencies worth addressing before/after merge.

### What's good
- The libultraship/spdlog logging guidance matches reality: SPDLOG_* macros are used across the codebase (games/mm/2s2h, combo/src, etc.), and `std::cerr` appears only in third-party `portable-file-dialogs.h`. The "no unconditional stderr in production" rule is already the de facto state.
- RAII + `std::optional<T>` for "not found" matches existing usage in `games/oot/soh/config/ConfigMigrators.h`, `N64WeirdFrames`, and `UIWidgets.hpp`.
- The three principles (no silent failures, context-appropriate mechanisms, predictable patterns) are a reasonable frame for a young combo codebase.
- Anti-patterns list is concrete and short — good.

### Issues

1. **C++23 `std::expected` conflicts with the project's C++20 baseline.** `README.md` line 27 states the build requirement is "C++20 compiler," and no file in the tree currently uses `std::expected`. Recommending it as a primary pattern sets up guidance that can't actually be followed without bumping the standard (or pulling in `tl::expected`). Either:
   - soften to "when C++23 is available, prefer `std::expected<T,E>`; until then, a custom `Result<T,E>` or `std::variant`-based return," or
   - make the C++ standard bump an explicit prerequisite and link it.

2. **"Blocks #31 (Cross-Game Context)" is stale.** Issue #31 is closed — the ComboContext/switching infrastructure already landed. The PR body should be updated to cite a currently-open issue this policy is supposed to gate, or drop the blocking claim. (The doc itself is still worth having; it's just the motivating link that's outdated.)

3. **Invented example identifiers should be flagged as illustrative.** `Combo_LoadGame`, `COMBO_SUCCESS`/`COMBO_ERR_INVALID_GAME`, and `ScopedLibrary` don't exist in the tree. That's fine for a policy doc, but a reader grepping for them will come up empty. Consider a one-line note (`Examples are illustrative; real APIs will follow these shapes.`) or swap in an actual existing function shape from `combo/` or `games/`.

4. **README "Documentation" section partially duplicates existing inline sections.** `README.md` already has a "Building" section (lines 23–45) and "Credits" section (lines 84–90), yet the new list links out to `docs/BUILDING.md` and `docs/CREDITS.md`. Not blocking, but either the inline sections should be trimmed to a pointer, or the duplicated links dropped. Also missing from the list: `docs/BUILDING_WINDOWS.md`, which is a natural companion to `BUILDING.md`.

### Gaps worth considering (follow-up, not blocking)
- No guidance on **assertions** (`assert`, `ZELDA_ASSERT`-style macros) vs. exceptions for invariant violations — libultraship and both games use assert-style macros heavily.
- No guidance on **error propagation across the C/C++ boundary** specifically for the OoT/MM decomp-derived C code calling into combo C++ (which is a major seam for this project).
- No guidance on **logging levels** (debug vs. info vs. warn vs. error) even though the doc says "gated by log level."
- Nothing on **user-facing vs. developer-facing errors** (fatal dialogs, crash reporting, save-file recovery) — arguably the most visible error paths for a randomizer.

### Suggestion
Land as a v1 after addressing (1) and (2); treat (3)/(4) and the gaps as a short follow-up to keep this PR scoped.

PR #221, opened 2026-04-19
fix: macOS libultraship linking with -force_load (#75)

### LINKED ISSUE #75: macOS build failing: OoT shared library not linking against libultraship
## Summary

The macOS build is failing on CI with undefined symbols for N64 OS functions. The OoT shared library is not properly linking against libultraship.

## Error

```
Undefined symbols for architecture x86_64:
  "_osCreateMesgQueue", referenced from: _OoT_bootproc in boot_main.c.o
  "_osRecvMesg", "_osStartThread", "_osViExtendVStart", etc.
```

## Root Cause

The OoT shared library (`soh.dylib`) needs to link against libultraship, but the symbols are not being resolved correctly. This needs investigation into the CMake configuration for macOS shared library builds.

## Current Status

- **Workaround**: macOS build job disabled in CI (`if: false`)
- **Linux/Windows**: Both building successfully

## Acceptance Criteria

- [ ] Identify why libultraship symbols aren't linking in the macOS OoT shared library
- [ ] Fix CMake configuration or linker flags
- [ ] Re-enable macOS build job in CI
- [ ] macOS CI build passes

## Related

This was discovered while working on PR #67 (Remove standalone build support).

spencerduncan:
## Summary

- Fix macOS link failures where N64 OS stubs (`osCreateMesgQueue`, `osRecvMesg`, `osStartThread`, etc.) from libultraship were dropped by the linker when building `soh.dylib` / `2ship.dylib`.
- Add `-force_load` for libultraship on macOS (`games/oot/CMakeLists.txt`, `games/mm/CMakeLists.txt`). Unlike Linux's `-Wl,-export-dynamic`, macOS's ld requires `-force_load` to pull in all symbols from a static archive into a shared library.
- Re-enable the `build-macos` job in `.github/workflows/generate-builds.yml`, which was previously gated on `if: false` as a workaround.

Fixes #75.

## Test plan

- [ ] CI `build-macos` job passes on this PR (previously disabled).
- [ ] CI `build-linux` and `build-windows` jobs still pass (change is behind `CMAKE_SYSTEM_NAME STREQUAL "Darwin"`).
- [ ] Resulting `SoH.dmg` artifact launches and loads both OoT and MM without undefined-symbol errors at dlopen time.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634623319.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634558013.zip)

### DIFF
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -110,10 +110,6 @@ jobs:
         path: deps
 
   build-macos:
-    # DISABLED: macOS build is currently broken - see GitHub issue for tracking
-    # Error: OoT shared library not linking against libultraship properly
-    # Re-enable once issue is resolved
-    if: false
     needs: generate-rsbs-otr
     runs-on: macos-14
     steps:
--- a/CMake/SingleExecutable.cmake
+++ b/CMake/SingleExecutable.cmake
@@ -178,6 +178,16 @@ if(UNIX AND NOT APPLE)
     target_link_options(redship PRIVATE -rdynamic)
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    # libultraship is a static archive containing the N64 OS stubs
+    # (osCreateMesgQueue, osRecvMesg, osStartThread, ...). ld64 drops archive
+    # members whose symbols look unreferenced; -force_load pulls them all in.
+    # Linux uses -Wl,-export-dynamic via -rdynamic above; macOS requires -force_load.
+    target_link_options(redship PRIVATE
+        "LINKER:-force_load,$<TARGET_FILE:libultraship>"
+    )
+endif()
+
 # ============================================================================
 # CTest Integration
 # ============================================================================
@@ -234,7 +244,13 @@ endif()
 # Installation
 # ============================================================================
 
-install(TARGETS redship RUNTIME DESTINATION . COMPONENT ship)
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    # CPack Bundle maps CMAKE_INSTALL_PREFIX to Contents/Resources; the executable
+    # belongs in Contents/MacOS for a valid .app bundle.
+    install(TARGETS redship RUNTIME DESTINATION ../MacOS COMPONENT ship)
+else()
+    install(TARGETS redship RUNTIME DESTINATION . COMPONENT ship)
+endif()
 
 message(STATUS "Single executable 'redship' will be built")
 message(STATUS "Use --test option to run integration tests")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -416,7 +416,11 @@ add_custom_target(CreateOSXIcons
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Creating OSX icons ..."
     )
-add_dependencies(soh CreateOSXIcons)
+if(SINGLE_EXECUTABLE_BUILD)
+    add_dependencies(redship CreateOSXIcons)
+else()
+    add_dependencies(soh CreateOSXIcons)
+endif()
 
 install(TARGETS ZAPD_OoT DESTINATION ${CMAKE_BINARY_DIR}/assets)
 
@@ -425,13 +429,20 @@ set(PROGRAM_PERMISSIONS_EXECUTE OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_EXECU
 install(DIRECTORY "${CMAKE_SOURCE_DIR}/games/oot/assets/extractor/" DESTINATION ./assets/)
 install(DIRECTORY "${CMAKE_SOURCE_DIR}/games/oot/assets/xml/" DESTINATION ./assets/xml)
 
-# Rename the installed soh binary to drop the macos suffix
-INSTALL(CODE "FILE(RENAME \${CMAKE_INSTALL_PREFIX}/../MacOS/soh-macos \${CMAKE_INSTALL_PREFIX}/../MacOS/soh)")
+if(SINGLE_EXECUTABLE_BUILD)
+    install(CODE "
+        include(BundleUtilities)
+        fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/../MacOS/redship\" \"\" \"${dirs}\")
+        ")
+else()
+    # Rename the installed soh binary to drop the macos suffix
+    INSTALL(CODE "FILE(RENAME \${CMAKE_INSTALL_PREFIX}/../MacOS/soh-macos \${CMAKE_INSTALL_PREFIX}/../MacOS/soh)")
 
-install(CODE "
-    include(BundleUtilities)
-    fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/../MacOS/soh\" \"\" \"${dirs}\")
-    ")
+    install(CODE "
+        include(BundleUtilities)
+        fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/../MacOS/soh\" \"\" \"${dirs}\")
+        ")
+endif()
 
 endif()
 
--- a/games/mm/CMakeLists.txt
+++ b/games/mm/CMakeLists.txt
@@ -849,7 +849,11 @@ execute_process(COMMAND ${CURL} -sSfL https://raw.githubusercontent.com/gabomdq/
 
 if("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
 configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/macosx/Info.plist.in ${CMAKE_BINARY_DIR}/macosx/Info.plist @ONLY)
-INSTALL(TARGETS 2ship DESTINATION ../MacOS COMPONENT 2s2h)
+if(NOT SINGLE_EXECUTABLE_BUILD)
+    # 2ship is a SHARED library in legacy mode; in single-exec mode it's OBJECT libs
+    # linked into redship, so there's nothing to install by that name.
+    INSTALL(TARGETS 2ship DESTINATION ../MacOS COMPONENT 2s2h)
+endif()
 INSTALL(FILES ${CMAKE_BINARY_DIR}/gamecontrollerdb.txt DESTINATION ../MacOS COMPONENT 2s2h)
 INSTALL(FILES ${CMAKE_BINARY_DIR}/mm/2ship.o2r DESTINATION ../Resources COMPONENT 2s2h)
 elseif(NOT "${CMAKE_SYSTEM_NAME}" MATCHES "NintendoSwitch|CafeOS")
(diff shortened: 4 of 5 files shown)


spencerduncan (commented):
## Review: APPROVE (pending build-macos green)

### Correctness
- Both CMake diffs sit inside `if (CMAKE_SYSTEM_NAME STREQUAL \"Darwin\")` blocks (oot `games/oot/CMakeLists.txt:572`, mm `games/mm/CMakeLists.txt:506`), so Linux/Windows/Switch/CafeOS paths are untouched. Linux keeps its existing `-Wl,-export-dynamic` at `games/oot/CMakeLists.txt:678` and `games/mm/CMakeLists.txt:612`.
- `\"LINKER:-force_load,$<TARGET_FILE:libultraship>\"` is the right idiom: the `LINKER:` prefix is the portable CMake form, and `$<TARGET_FILE:libultraship>` resolves at build time to the actual archive path. libultraship is already an `add_dependencies` target (`games/oot/CMakeLists.txt:713`) and linked via `target_link_libraries` (`games/oot/CMakeLists.txt:263`), so the generator expression will resolve correctly.
- Workflow un-gate is a clean revert of the `if: false` workaround — no other job references need updating.

### Is `-force_load` the right approach?
Yes. On macOS `ld64`, there is no `-export-dynamic`; to pull unreferenced symbols (the `osCreateMesgQueue` / `osRecvMesg` N64 OS stubs from the static archive) into `soh.dylib` / `2ship.dylib`, the options are `-all_load` (forces every archive, blunt) or `-force_load <archive>` (targeted). `-force_load` is the correct, minimal choice and matches what projects like LLVM/Swift do. It assumes libultraship builds as a STATIC archive on macOS — the symptom in issue #75 confirms that.

### Risks
- **build-macos is still `pending`** at review time — other jobs (linux/windows/clang-format/clang-tidy/integration) are all green. The real validation is that pending job; merge should wait for it. The test plan checklist in the PR description is also unchecked.
- Minor binary-size growth on macOS from pulling the full libultraship archive into each `.dylib`. Unavoidable given the approach; acceptable.
- Low risk of duplicate-symbol errors if libultraship ever gains weakly-defined symbols that collide across both game dylibs, but that would surface as a build failure, not a silent issue.

### Nits (non-blocking)
- The comment block is duplicated verbatim in both CMakeLists.txt. Fine given the files are parallel; a shared helper function would be over-engineering for two call sites.
- `TARGET_FILE` requires libultraship to be a real build target (STATIC/SHARED/EXECUTABLE) — which it is via `add_subdirectory` at `games/oot/CMakeLists.txt:96`. If it were ever refactored to INTERFACE or OBJECT, this flag would need revisiting.

**Verdict: APPROVE** contingent on the build-macos job passing. If build-macos fails, that's the signal that libultraship is built as something other than a static archive on macOS and this needs revisiting.

spencerduncan:
### macOS config failure — fixed in d953dc9

The rebased macOS job was failing at CMake **configure**, not link. Root cause was that `SINGLE_EXECUTABLE_BUILD=ON` is the default, so `soh` and `2ship` are OBJECT libraries linked into `redship` — not installable targets — but the Darwin-specific install blocks still referenced them by name:

```
CMake Error at games/oot/CMakeLists.txt:831 (INSTALL):
  INSTALL TARGETS given target "soh" which does not exist.
CMake Error at games/mm/CMakeLists.txt:824 (INSTALL):
  INSTALL TARGETS given target "2ship" which does not exist.
CMake Error at CMakeLists.txt:419 (add_dependencies):
  Cannot add target-level dependencies to non-existent target "soh".
```

The Linux and Windows install blocks already had `NOT SINGLE_EXECUTABLE_BUILD` guards; the Darwin blocks did not. Changes:

- `games/oot/CMakeLists.txt`, `games/mm/CMakeLists.txt`: gate `INSTALL(TARGETS soh|2ship ...)` on `NOT SINGLE_EXECUTABLE_BUILD` (matching the existing Linux/Windows pattern).
- `CMakeLists.txt`: branch `add_dependencies(... CreateOSXIcons)` and `fixup_bundle` on the build mode (`redship` in single-exec, `soh` in legacy). The legacy `soh-macos` rename is gated off in single-exec mode.
- `CMake/SingleExecutable.cmake`: on Darwin, install `redship` into `../MacOS` (Contents/MacOS) instead of `.` (Contents/Resources) so the .app bundle is valid.
- `CMake/SingleExecutable.cmake`: add `-force_load` to the `redship` target on Darwin. The PR's existing `target_link_options` in the `foreach(_t ${SOH_COMPILE_TARGETS})` loop is a no-op in single-exec mode (OBJECT libs don't link), so without this the `redship` link would have dropped N64 OS stubs for the same reason the soh.dylib link did.

spencerduncan (commented):

# Code Review — PR #221: fix: macOS libultraship linking with -force_load (#75)

**Repo:** spencerduncan/redshipblueship
**Branch:** `claude/fix-macos-libultraship-linking-sARn2` → `main`
**Commits:** 2 (a36e812, d953dc9)
**Diff:** +55 −14 across 5 files

**Verdict: REQUEST CHANGES** — the code is correct in principle, but the latest `build-macos` CI run (generate-builds #624, after the d953dc9 fix) still failed. The PR's own test plan ("CI build-macos job passes on this PR") is not satisfied.

---

## 1. CMake correctness — OK

All three force_load sites are properly guarded:

- `CMake/SingleExecutable.cmake` L180–188: inside `if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")`
- `games/oot/CMakeLists.txt` L595–605: inside the Darwin branch of the `GNU|Clang|AppleClang` compiler block
- `games/mm/CMakeLists.txt` L529–539: same pattern

Linux/Windows/Switch/CafeOS paths are untouched. The generator expression form `"LINKER:-force_load,$<TARGET_FILE:libultraship>"` is the portable CMake idiom — `LINKER:` handles driver-prefix translation, and `$<TARGET_FILE:libultraship>` resolves at build time. libultraship is added via `add_subdirectory` and linked via `target_link_libraries`, so the target exists and the generator expression resolves correctly.

The workflow un-gate (`.github/workflows/generate-builds.yml`) is a clean removal of the `if: false` workaround — no dependent job references need updating.

## 2. Static library selection — minimal and correct

Only `libultraship` is force-loaded, not the whole link set. This is the minimal choice: libultraship is the specific static archive containing the N64 OS stubs (`osCreateMesgQueue`, `osRecvMesg`, `osStartThread`) that ld64 drops due to dead-code elimination at the static-archive-into-shared-library boundary. Binary-size growth is bounded (just libultraship, pulled once per dylib / once into redship) and acceptable.

`-force_load` is the right tool vs the alternatives:
- `-all_load` — too blunt, would pull every archive
- `-Wl,-export-dynamic` — Linux-only, no ld64 equivalent
- `--whole-archive` / `--no-whole-archive` — GNU ld syntax, not ld64

This matches what LLVM/Swift and similar projects do for the same problem.

## 3. Symbol visibility / root cause — properly addressed

The fix addresses the actual root cause, not a symptom. The PR description and commit message correctly describe the mechanism: on macOS, when a static archive is linked into a shared library, ld64 only pulls archive members that resolve currently-unresolved references. The N64 OS stubs are referenced indirectly (via dispatch) from OoT/MM code paths that look unreferenced from the static archive's perspective at link time, so they get silently dropped and fail at dlopen with `_osCreateMesgQueue` undefined.

The second commit (d953dc9) adds a second necessary piece: force_load is also applied to `redship` directly in SingleExecutable.cmake. This matters because in single-executable mode, `soh` and `2ship` are OBJECT libs (not real link steps), so `target_link_options` on them is a no-op — only the final redship link sees force_load. Good catch.

## 4. CI coverage — BLOCKING

This is the blocking issue.

- `build-macos` status on PR #221 checks: **failed**
- Latest `generate-builds #624` (triggered by synchronize of commit d953dc9): **failed**
- clang-format, Static Analysis, integration-tests, build-linux, build-windows: green

The author's self-review comment (Apr 20) was "APPROVE pending build-macos green." The second commit (d953dc9) was specifically added to fix the SINGLE_EXECUTABLE_BUILD config errors that were blocking macOS configure. But per the post-synchronize workflow run, the build is still failing — meaning either:

(a) there's another macOS-specific config/link error beyond what d953dc9 addressed,
(b) the force_load fix itself is incomplete (e.g., libultraship isn't a STATIC archive on macOS, or there's a second archive that also needs force_load), or
(c) a transient runner/cache issue — but that's the less likely explanation given the consistent failure across two runs.

Without visibility into the failing log (requires GitHub sign-in), I can't disambiguate. **The test plan item "CI build-macos job passes on this PR (previously disabled)" is unchecked and the CI signal contradicts it.** This needs to be resolved before merge.

## 5. Comparison to upstream

Upstream HarbourMasters/Shipwright does **not** use `-force_load`. Searched `soh/CMakeLists.txt` on `develop`: the Darwin `target_link_options` only passes `-pthread`. Upstream also has `-Wl,-export-dynamic` on Linux (line 589) — matching the pattern this PR cites.

The reason upstream doesn't need force_load: upstream builds `soh` as an executable, not as a shared library that gets loaded by a wrapper. redshipblueship's architecture (soh/2ship as `.dylib` or as OBJECT libs linked into a `redship` wrapper) is what creates the static-archive-into-shared-lib boundary that triggers ld64's dead-code elimination. So the fix is fork-specific, not a missing upstream change. The approach is still sound.

## Nits (non-blocking)

- The `-force_load` comment block is duplicated verbatim across `games/oot/CMakeLists.txt` and `games/mm/CMakeLists.txt`. Fine given the files are parallel; a shared helper would be over-engineering.
- `$<TARGET_FILE:libultraship>` assumes libultraship resolves to a real linkable file (STATIC/SHARED/EXECUTABLE). If it's ever refactored to INTERFACE or OBJECT, this flag needs revisiting. Worth a comment in-tree if there's any chance of that refactor.
- The second commit (d953dc9) does more than its title suggests — it restructures the Darwin install blocks substantially (gating on `NOT SINGLE_EXECUTABLE_BUILD`, adding parallel paths for redship, moving install destination to `../MacOS` for valid `.app` structure). This is correct, but worth calling out in the PR description that the scope grew beyond "macOS linking."
- The "Build Artifacts" section in the PR description shows only rsbs-windows.zip and rsbs-linux.zip — no macOS artifact, consistent with the build-macos failure.

## Recommendation

**REQUEST CHANGES.** Resolve the build-macos CI failure before merging. The code is well-structured and the approach is correct — but the whole point of this PR is "make macOS build," and right now it doesn't. Once the log is visible, the next step is either a targeted fix for whatever ld/configure error remains, or a confirmation that the force_load approach needs to be broader (additional archives, or switching to something like `-Wl,-all_load` scoped per-target).

If the current failure turns out to be a flake, re-run and confirm green before approving — but don't merge on an unverified flake assumption.

spencerduncan:
Closing for now — macOS isn't currently an active build target (it's SKIPPED on all green main-branch CI runs), and this PR's own macOS build is still failing after two fix attempts. Reopen when macOS becomes a real target and we have a working local repro.

Issue #177, opened 2026-01-31
Investigate Windows CI build time (10+ min vs Linux 2 min)


spencerduncan:
## Problem

Windows builds in CI take 10+ minutes while Linux builds complete in ~2 minutes. This significant difference suggests optimization opportunities.

## Potential causes to investigate

1. **Cache effectiveness**
   - Linux uses ccache via `hendrikmuhs/ccache-action`
   - Windows uses sccache - is it hitting cache as effectively?
   - Check cache hit rates in CI logs

2. **vcpkg dependency installation**
   - Windows installs vcpkg packages every run
   - Are these being cached properly?
   - Compare with Linux which uses a Docker image with pre-installed deps

3. **Compiler performance**
   - MSVC vs GCC/Clang compilation speed differences
   - Debug info generation (`/Zi` vs `/Z7`) overhead
   - Link-time code generation (`/LTCG`) if enabled

4. **Linker performance**
   - MSVC linker is notoriously slower than ld/lld
   - Consider using lld-link on Windows
   - Response file handling overhead

5. **I/O differences**
   - GitHub Actions Windows runners may have slower I/O
   - Antivirus scanning overhead on Windows runners

## Data to collect

- [ ] Cache hit/miss rates for sccache on Windows
- [ ] Time breakdown by build phase (configure, compile, link)
- [ ] vcpkg install time
- [ ] Compare with/without `/LTCG`

## Related

- #149 disabled IPO for OTRExporter/ZAPDTR in single-exe mode
- Linux uses Docker image with pre-cached dependencies

spencerduncan:
## CI gap worth folding into this issue: `build-windows` runs **no** ctest at all

Found during a bugbash sweep, verified against `origin/main` today. Not a request to change anything here — recording it because it lands on the same job this issue is about.

### (a) The gap

`build-windows` (`.github/workflows/generate-builds.yml:454`, `runs-on: windows-latest` at `:472`) has exactly these steps, `:479` through `:579`:

```
Export GitHub Actions cache env vars for vcpkg   :479
Install dependencies                             :485
Git Checkout                                     :489
Configure sccache                                :495
Restore Cached VCPKG folder                      :521
Configure Developer Command Prompt               :530
Download soh.o2r                                 :532
Download 2ship.o2r                               :537
CMake configure (Windows)                        :547
Zero sccache stats                               :557
CMake build (Windows)                            :559
sccache stats (Windows)                          :563
Package (Windows)                                :566
[...]

PR #224, opened 2026-04-20
ci: optimize Windows build time (#177)

spencerduncan:
## Summary

Windows CI runs are dominated by vcpkg compiling static deps (sdl2, glew, libzip, boost, etc.) from source on every run. This enables vcpkg's **GitHub Actions binary cache provider** so unchanged ports are fetched pre-built via the GHA cache instead of rebuilt. Also widens the sccache cap so MSVC object caching survives across jobs.

Fixes issue #177.

## Changes (`.github/workflows/generate-builds.yml`, `build-windows` job only)

- **vcpkg binary caching**: job-level `VCPKG_BINARY_SOURCES=clear;x-gha,readwrite`, plus `actions/github-script@v7` step to export `ACTIONS_CACHE_URL` / `ACTIONS_RUNTIME_TOKEN` (required — vcpkg runs inside CMake, not an action step, so these have to be re-exported to the job env).
- **sccache**: bump `max-size` 2G → 5G and drop `evict-old-files: job` so stored objects persist across jobs (sccache's own LRU still enforces the cap).
- **vcpkg folder cache**: key bumped to `v2-` and now hashes `CMake/automate-vcpkg.cmake` in addition to `CMakeLists.txt` (so bootstrap-script changes invalidate it); old `v1-` key kept as a restore fallback so the first run after merge still warms from the existing cache.

No changes to application code, submodules, or other workflow jobs.

## Expected time savings

- **vcpkg binary cache hit** (steady state on a PR branch or post-warmup on `main`): ~10–20 min removed from the run — `vcpkg install` goes from compiling every port from source to unpacking prebuilt binaries.
- **sccache cap bump**: fewer evictions → higher hit rate on deep edits, expected additional 1–3 min saved on typical PRs.
- **First run on `main` after merge**: no speed-up (cold cache warms); **subsequent runs**: the big win lands.

## Test plan

- [ ] CI runs green on this PR; confirm the `build-windows` job still produces `rsbs-windows.zip`.
- [ ] Check `build-windows` logs for lines like `Restored from GHA cache` / `Stored in GHA cache` during the vcpkg install phase — confirms the x-gha provider is wired.
- [ ] Compare `build-windows` duration against a recent `main` run (before) and a re-run on this branch once the cache is warm (after).
- [ ] Verify sccache stats line near end of build shows non-zero hit rate on re-runs.
- [ ] Confirm the produced `rsbs-windows` artifact is identical in contents to pre-change builds (no dependency ABI drift from the binary cache).

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6522371827.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6522275314.zip)

### DIFF
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -327,7 +327,18 @@ jobs:
   build-windows:
     needs: generate-rsbs-otr
     runs-on: windows-latest
+    env:
+      # vcpkg binary caching via GitHub Actions cache. vcpkg stores prebuilt port
+      # binaries keyed by their ABI hash, so unchanged deps skip source compilation.
+      # Requires ACTIONS_CACHE_URL and ACTIONS_RUNTIME_TOKEN, exported below.
+      VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
     steps:
+    - name: Export GitHub Actions cache env vars for vcpkg
+      uses: actions/github-script@v7
+      with:
+        script: |
+          core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+          core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
     - name: Install dependencies
       run: |
         choco install ninja llvm -y
@@ -342,8 +353,7 @@ jobs:
       uses: hendrikmuhs/ccache-action@v1.2
       with:
         variant: sccache
-        max-size: "2G"
-        evict-old-files: job
+        max-size: "5G"
         save: ${{ github.ref_name == github.event.repository.default_branch }}
         key: ${{ runner.os }}-sccache-${{ github.ref }}
         restore-keys: |
@@ -354,8 +364,9 @@ jobs:
       id: restore-cache-vcpkg
       uses: actions/cache/restore@v4
       with:
-        key: ${{ runner.os }}-vcpkg-v1-${{ hashFiles('CMakeLists.txt') }}
+        key: ${{ runner.os }}-vcpkg-v2-${{ hashFiles('CMakeLists.txt', 'CMake/automate-vcpkg.cmake') }}
         restore-keys: |
+          ${{ runner.os }}-vcpkg-v2-
           ${{ runner.os }}-vcpkg-v1-
         path: vcpkg
     - name: Configure Developer Command Prompt
@@ -387,5 +398,5 @@ jobs:
       if: ${{ github.ref_name == github.event.repository.default_branch }}
       uses: actions/cache/save@v4
       with:
-        key: ${{ runner.os }}-vcpkg-v1-${{ hashFiles('CMakeLists.txt') }}
+        key: ${{ runner.os }}-vcpkg-v2-${{ hashFiles('CMakeLists.txt', 'CMake/automate-vcpkg.cmake') }}
         path: vcpkg


spencerduncan:

spencerduncan (commented):
## Review — `ci: optimize Windows build time` (#177)

**Verdict: APPROVE with one suggested follow-up.**

The pattern here is canonical (matches the Microsoft vcpkg docs for the `x-gha` binary backend) and the changes are well-scoped to the `build-windows` job. Caching gains look realistic — vcpkg port compilation genuinely is the long pole on Windows, and binary unpacking should reclaim most of it on warm runs.

### Correctness of the GHA cache integration
- Re-exporting `ACTIONS_CACHE_URL` / `ACTIONS_RUNTIME_TOKEN` via `actions/github-script` is the right approach: vcpkg is invoked from CMake (via `automate-vcpkg.cmake` → `vcpkg install`), not as a GHA step, so it doesn't inherit the runtime context that `actions/cache` gets implicitly. The `|| ''` defaults are also correct — fork PRs won't have `ACTIONS_RUNTIME_TOKEN`, and vcpkg will degrade gracefully to no-cache instead of erroring.

- ⚠️ **One real concern**: GitHub deprecated the v1 cache service in early 2025 and migrated `actions/cache@v4` to a v2 backend that uses `ACTIONS_RESULTS_URL` (not `ACTIONS_CACHE_URL`). Recent vcpkg revisions added support for the v2 API and prefer `ACTIONS_RESULTS_URL` when present. Since `CMake/automate-vcpkg.cmake` does `git clone --depth 1` with no pin and `git pull` on cache restore, you're always on vcpkg HEAD — so the v2-aware code path will be hit. **Suggest also exporting `ACTIONS_RESULTS_URL`** in the same script step:
  ```yaml
  core.exportVariable('ACTIONS_RESULTS_URL', process.env.ACTIONS_RESULTS_URL || '');
  ```
  Without this, the x-gha provider may silently no-op on newer vcpkg, and the run will look fine while quietly rebuilding everything from source. The Test plan item about checking for `Restored from GHA cache` log lines is the right way to validate, but worth pre-empting.

### Security implications of `ACTIONS_CACHE_URL` / `ACTIONS_RUNTIME_TOKEN` exposure
- Net-zero increase in blast radius. Both vars are already implicitly populated for `actions/cache/restore@v4` and `hendrikmuhs/ccache-action@v1.2`, which run in the same job. Promoting them to job-level env just makes them visible to subprocesses — which is exactly the intent. The token is per-run, expires in ~6h, and is scoped to that run's cache namespace.
- Fork-PR safety holds: GitHub strips `ACTIONS_RUNTIME_TOKEN` for `pull_request` events from forks, so cache writes from forks aren't possible regardless of `,readwrite`.

### Cache poisoning / stale builds
- **vcpkg x-gha**: keyed by vcpkg's internal ABI hash (compiler version, port source, dep graph), so a stale binary requires an ABI-hash collision — not a realistic concern. Same trust boundary as any other write to the repo cache.
- **sccache cap 2G → 5G + dropping `evict-old-files: job`**: safe. sccache's content-addressed keys mean stale objects are unreachable, not incorrect; the LRU enforces the 5G cap. Removing the per-job eviction increases hit rate on deep edits, as advertised.
- **`v1-` restore-keys fallback**: safe. The restored `vcpkg/` folder is just the checkout dir; `_install_or_update_vcpkg()` in `automate-vcpkg.cmake` does `git pull` on existing dirs, so the toolchain self-updates. Any port whose ABI hash drifted will be rebuilt (or fetched from x-gha) — no ABI mixing risk.
- The `save` gate on `github.ref_name == default_branch` for both sccache and the vcpkg folder cache is correct cache hygiene — PR runs read from main's cache but don't pollute it.

### Estimated real-world impact
The 10–20 min vcpkg estimate looks credible — Windows cold vcpkg installs typically run 15–25 min; binary unpack drops to <1 min. The +1–3 min from the sccache bump is plausible but I'd expect it closer to the low end on most PRs (sccache hits dominate either way; cap mostly matters for big diffs that touch many headers). First post-merge `main` run gets nothing; subsequent runs land the win. The CI status here (build-windows still pending) won't actually validate the steady-state benefit until a re-run on a warm cache — worth doing before the test-plan checkboxes are ticked off.

### Minor
- Pre-existing, not introduced here: `runs-on: windows-latest` floats, so a runner image bump can silently invalidate caches independent of the key.
- Nit: typo in this task's instructions referenced `gh pr review 221`; posting on 224 as the actual PR.

Issue #168, opened 2026-01-29
MM first-entry SaveContext initialization for cross-game switch


spencerduncan:
## Problem

When switching from OoT to MM for the first time, there's no frozen MM state to restore. MM needs a valid initial SaveContext with:
1. The startup entrance injected (Clock Tower Interior from Mido's House link)
2. Randomizer-compatible initial state (skip cycle zero, proper flags set)
3. **MM must be in randomizer mode** (matching OoT's rando state)

## Current behavior

The startup entrance injection exists in `z_play.c`:
```c
uint16_t startupEntrance = Combo_GetStartupEntrance();
if (startupEntrance != 0) {
    gSaveContext.save.entrance = startupEntrance;
    Combo_ClearStartupEntrance();
}
```

But it's unclear what state `gSaveContext` is in at this point for a first-time entry, and whether MM knows it should be in randomizer mode.

## Expected behavior

On first cross-game switch to MM:
- MM should initialize with a **new randomizer SaveContext** (not vanilla new game)
- This means: cycle zero skipped, appropriate randomizer flags set
- **MM must be in randomizer mode** — `IS_RANDO` should be true
- Startup entrance (0xC010 Clock Tower Interior) should be respected

## Approach

- Trace MM's save initialization path in randomizer mode
- Propagate OoT's randomizer state to MM during cross-game switch
- Ensure `MM_Game_Init` sets up a rando-compatible SaveContext before `z_play.c` runs
- Verify startup entrance is applied after SaveContext init, not before

## Test plan

- [ ] Boot OoT in rando mode → enter Mido's House → verify MM starts at Clock Tower
- [ ] Verify MM `IS_RANDO` is true
- [ ] Verify MM is in randomizer state (not vanilla cycle zero)
- [ ] Verify Link's position/state is sensible

## Context

Part of: #145 (single-exe architecture)
Depends on: #153 (lifecycle), #165 (MM audio fix)

PR #240, opened 2026-04-20
feat: OoT state restoration on return from MM (#170)

### LINKED ISSUE #170: OoT state restoration on return from MM (roundtrip)
## Problem

The return path (MM → OoT) after a cross-game switch hasn't been tested. When the player exits Clock Tower in MM, they should return to OoT at Mido's House with their OoT state fully restored.

## Current implementation

1. OoT state is frozen via `Context_FreezeState()` during switch to MM
2. Return entrance (0x0443 Kokiri from Mido's) is stored with frozen state
3. When MM exits Clock Tower (entrance 0xD800), reverse link triggers switch back
4. `Context_RestoreState()` should restore OoT's SaveContext
5. OoT resumes at the return entrance

## Risk scenarios

1. OoT SaveContext not properly restored (inventory, flags, position)
2. OoT audio/graphics state not properly resumed after suspend
3. Return entrance not applied correctly
4. OoT thinks it's a fresh boot instead of resume

## Test plan

- [ ] Full roundtrip: OoT (rando) → Mido's House → MM Clock Tower → exit → OoT
- [ ] Verify OoT inventory/flags preserved across roundtrip
- [ ] Verify Link spawns at correct position (outside Mido's House)
- [ ] Verify OoT audio/graphics resume without glitches
- [ ] Verify multiple roundtrips work (OoT → MM → OoT → MM → OoT)

## Context

Part of: #145 (single-exe architecture)
Depends on: #153 (lifecycle), #166 (OoT suspend/resume)
Related: #168 (MM first-entry), entrance wiring in #151

spencerduncan:
## Summary

Fixes the MM→OoT return leg of the cross-game roundtrip. Before this PR, the return path was silently broken — switching from OoT into MM worked, but coming back left OoT's SaveContext corrupted because the MM→OoT trigger never fired and MM's SaveContext was never frozen.

Closes #170.

## Root cause

`Combo_CheckEntranceSwitch` is defined in OoT's port layer (`games/oot/soh/GameExports_SingleExe.cpp`) and, in single-exe mode, resolves the `extern` references from both OoT's `z_play.c` and MM's `z_play.c` / `z_player.c`. The implementation hardcoded `\"oot\"` as the source game — so when MM called it with MM entrance id `0xD800` (South Clock Town, spawn 0), the function looked up OoT's entrance table for that id, found nothing, and returned without ever freezing state or setting up the pending switch.

Compounding the problem, no `Context_RestoreState` call ever ran on the resume side: `OoT_Game_Resume` only re-init'd audio, so even when the freeze *was* captured, the unified `gSaveContext` storage (see `src/common/unified_save.c` — both games share one buffer interpreted through different struct layouts) stayed populated with whatever MM had written, not OoT's pre-switch state.

## Architecture

\`\`\`
                 OoT                                  MM
                  │                                    │
   ┌──────────────┴──────────────┐      ┌─────────────┴──────────────┐
   │ z_play.c                    │      │ z_play.c                   │
   │   Combo_CheckEntranceSwitch │ ◄── same symbol ──► Combo_CheckEntranceSwitch │
   └──────────────┬──────────────┘      └─────────────┬──────────────┘
                  │                                    │
                  ▼                                    ▼
           dispatches on Context_GetCurrentGame() ──► \"oot\" | \"mm\"
                              │
                              ▼
                   Combo_CheckCrossGameEntrance(gameId, ...)
                              │
                              ▼
                     sets gPendingSwitch
                              │
                              ▼
            Combo_FreezeState(gameId, &gSaveContext, ...) ──► Context_FreezeState
                              │
                              ▼
                   main.cpp detects pending switch
                              │
                              ▼
                   GameRunner_SwitchTo(target)
                              │
                              ▼
              OoT_Game_Suspend / MM_Game_Suspend
                              │
                              ▼
              OoT_Game_Resume / MM_Game_Resume  ◄── NEW:
                              │                     Context_RestoreState(...)
                              ▼                     + gSaveContext.entrance = target
                              │
                   ops->run() — game resumes with correct SaveContext
\`\`\`

## Changes

- **\`games/oot/soh/GameExports_SingleExe.cpp\`**
  - \`Combo_CheckEntranceSwitch\` now dispatches on \`Context_GetCurrentGame()\` so MM-side calls freeze MM state against MM's entrance table (and vice versa). Early-boot default is \`\"oot\"\` since OoT is always the first game.
  - Re-entrancy guard: if a cross-game switch is already queued, skip re-freezing to avoid capturing mutated state from a death sequence or mid-transition save.
  - \`OoT_Game_Resume\` now calls \`Context_RestoreState(GAME_OOT, ...)\` when a frozen state exists, and applies the correct entrance (startup entrance preferred, frozen return entrance as fallback).
- **\`games/mm/2s2h/GameExports_SingleExe.cpp\`**
  - Mirror fix in \`MM_Game_Resume\`: restore MM's frozen SaveContext and entrance on every resume after the first. First boot still falls through to the existing \`MM_InitFirstEntrySaveContext\` path from #168.
- **\`rsbs/src/main.cpp\`**
  - Call \`Context_SetCurrentGame(...)\` after \`GameRunner_StartGame\` and every successful \`GameRunner_SwitchTo\` so the dispatch in \`Combo_CheckEntranceSwitch\` resolves to the right side.
- **Tests**
  - Expand the CTest \`Roundtrip\` test (\`src/common/test_runner.cpp\`) to drive the full OoT→MM→OoT round-trip through the production \`Combo_*\` C API — not the internal C++ \`Entrance_*\` API that the old test used, which would not have caught this bug — and assert both SaveContexts survive intact.
  - Add \`ReturnPathTest\` gtest suite (\`combo/tests/context_test.cpp\`) covering: MM-side return detection, OoT-side negative case for MM entrance ids, full round-trip fingerprinting, and restore-without-freeze safety.

## Scope notes

- **Entrance codes verified** against the actual OoT/MM source: \`0x0433\` = \`ENTR_MIDOS_HOUSE_0\` (into), \`0x0443\` = \`ENTR_KOKIRI_FOREST_OUTSIDE_MIDOS_HOUSE\` (out of), \`0xD800\` = \`ENTRANCE(SOUTH_CLOCK_TOWN, 0)\`, \`0xC010\` = \`ENTRANCE(CLOCK_TOWER_INTERIOR, 1)\`. Existing \`OOT_ENTR_KOKIRI_FROM_MIDOS\` / \`MM_ENTR_SOUTH_CLOCK_TOWN_0\` names in \`src/common/entrance.h\` are already correct.
- **Not touched**: submodules, \`.github/workflows/*\`, root \`CMakeLists.txt\`.
- **Not fixed here**: the \`OOT_SAVE_CONTEXT_SIZE\` mismatch between \`src/common/game.h\` (\`0x1428\`, N64 original) and \`src/common/unified_save.c\` (\`0x22000\`, current SoH layout with SohStats). \`Context_FreezeState\` currently clamps to \`0x1428\`, so only the N64 core is round-tripped — extended SoH/2SH fields aren't preserved. Worth a follow-up issue; out of scope here.

## Test plan

- [ ] CI: \`build-windows\`, \`build-linux\`, \`build-macos\`, \`clang-format\` must pass
- [ ] CTest \`Roundtrip\` target passes (new assertions exercise MM-side detection)
- [ ] CTest \`SwitchOoTMM\`, \`SwitchMMOoT\`, \`Context\`, \`AllTests\` still pass
- [ ] Manual: boot OoT → Mido's House → ends up in MM Clock Tower Interior → exit to South Clock Town → returns to OoT with Link back in Kokiri Forest (outside Mido's) and inventory intact
- [ ] Manual: repeat round-trip multiple times; SaveContext fingerprints (health, rupees, equipment, entrance) remain stable

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634668501.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634550315.zip)

### DIFF
--- a/combo/tests/context_test.cpp
+++ b/combo/tests/context_test.cpp
@@ -420,3 +420,113 @@ TEST(IntegrationTest, ReturnSwitchToMMRestoresFrozenState) {
     // Cleanup
     Context_ClearAllFrozenStates();
 }
+
+// ============================================================================
+// Return-path regression tests for issue #170
+// ============================================================================
+//
+// Before #170, Combo_CheckEntranceSwitch hardcoded "oot" as the source game
+// even when invoked from MM's z_play.c. That meant MM→OoT transitions never
+// resolved a matching cross-game link, and MM's SaveContext was never frozen.
+// These tests lock in the fix by exercising the C API that the game code
+// actually calls, asserting it works from both sides.
+
+TEST(ReturnPathTest, MMSideCheckFindsReturnEntrance) {
+    Context_Init();
+    Entrance_Init();
+    Entrance_RegisterTestLinks();
+
+    // Simulate MM stepping onto the return entrance (South Clock Town spawn 0).
+    // Using the C API with an explicit "mm" game id — this is the pathway the
+    // production fix takes via Context_GetCurrentGame() dispatch.
+    Combo_CheckCrossGameEntrance("mm", MM_ENTR_SOUTH_CLOCK_TOWN_0);
+
+    EXPECT_TRUE(Combo_IsCrossGameSwitch());
+    EXPECT_STREQ(Combo_GetSwitchTargetGameId(), "oot");
+    EXPECT_EQ(Combo_GetSwitchTargetEntrance(), OOT_ENTR_KOKIRI_FROM_MIDOS);
+
+    Entrance_ClearPendingSwitch();
+    Entrance_ClearLinks();
+}
+
+TEST(ReturnPathTest, OoTSideCheckDoesNotMatchMMEntrance) {
+    // Regression guard: the buggy implementation checked "oot" links for MM
+    // entrance 0xD800, which would silently match nothing. Verify that when
+    // the right game id ("oot") is paired with an MM entrance, no switch
+    // triggers — that's the correct negative behavior.
+    Context_Init();
+    Entrance_Init();
+    Entrance_RegisterTestLinks();
+
+    Combo_CheckCrossGameEntrance("oot", MM_ENTR_SOUTH_CLOCK_TOWN_0);
+    EXPECT_FALSE(Combo_IsCrossGameSwitch());
+
+    Entrance_ClearPendingSwitch();
+    Entrance_ClearLinks();
+}
+
+TEST(ReturnPathTest, FullRoundtripPreservesOoTSaveContext) {
+    Context_Init();
+    Entrance_Init();
+    Entrance_RegisterTestLinks();
+
+    // --- Leg 1: OoT → MM ---
+    // Populate OoT SaveContext with a recognizable fingerprint.
+    uint8_t ootSave[OOT_SAVE_CONTEXT_SIZE];
+    memset(ootSave, 0x11, sizeof(ootSave));
+    ootSave[0] = 'O'; ootSave[1] = 'o'; ootSave[2] = 'T'; ootSave[3] = '!';
+
+    // OoT's entrance hook fires while OoT is current.
+    Context_SetCurrentGame(GAME_OOT);
+    Combo_CheckCrossGameEntrance("oot", OOT_ENTR_MIDOS_HOUSE);
+    EXPECT_TRUE(Combo_IsCrossGameSwitch());
+    uint16_t ootReturn = Combo_GetSwitchReturnEntrance();
+    Combo_FreezeState("oot", ootReturn, ootSave, sizeof(ootSave));
+    Entrance_ClearPendingSwitch();
+
+    // --- Leg 2: pretend to run MM and freeze MM state ---
+    Context_SetCurrentGame(GAME_MM);
+    uint8_t mmSave[MM_SAVE_CONTEXT_SIZE];
+    memset(mmSave, 0x22, sizeof(mmSave));
+    mmSave[0] = 'M'; mmSave[1] = 'M';
+
+    Combo_CheckCrossGameEntrance("mm", MM_ENTR_SOUTH_CLOCK_TOWN_0);
+    EXPECT_TRUE(Combo_IsCrossGameSwitch());
+    EXPECT_STREQ(Combo_GetSwitchTargetGameId(), "oot");
+    uint16_t mmReturn = Combo_GetSwitchReturnEntrance();
+    Combo_FreezeState("mm", mmReturn, mmSave, sizeof(mmSave));
+    Entrance_ClearPendingSwitch();
+
+    // --- Leg 3: restore OoT on return, verify data integrity ---
+    uint8_t restoredOoT[OOT_SAVE_CONTEXT_SIZE] = {0};
+    int result = Combo_RestoreState("oot", restoredOoT, sizeof(restoredOoT));
+    EXPECT_EQ(result, 1);
+    EXPECT_EQ(restoredOoT[0], 'O');
+    EXPECT_EQ(restoredOoT[1], 'o');
+    EXPECT_EQ(restoredOoT[2], 'T');
+    EXPECT_EQ(restoredOoT[3], '!');
+    EXPECT_EQ(Combo_GetFrozenReturnEntrance("oot"), OOT_ENTR_KOKIRI_FROM_MIDOS);
+
+    // MM's frozen state still sits alongside OoT's — both coexist.
+    uint8_t restoredMM[MM_SAVE_CONTEXT_SIZE] = {0};
+    EXPECT_EQ(Combo_RestoreState("mm", restoredMM, sizeof(restoredMM)), 1);
+    EXPECT_EQ(restoredMM[0], 'M');
+
+    // Cleanup
+    Context_ClearAllFrozenStates();
+    Entrance_ClearLinks();
+}
+
+TEST(ReturnPathTest, RestoreWithoutFreezeIsSafe) {
+    // Edge case: first boot of a game — OoT_Game_Resume could be reached
+    // without any prior freeze (e.g. if GameRunner misbehaves). The restore
+    // must refuse rather than memcpy zeros over a live SaveContext.
+    Context_Init();
+    Context_ClearAllFrozenStates();
+
+    uint8_t buffer[OOT_SAVE_CONTEXT_SIZE];
+    memset(buffer, 0xFF, sizeof(buffer));
+    int result = Context_RestoreState(GAME_OOT, buffer, sizeof(buffer));
+    EXPECT_EQ(result, 0);
+    EXPECT_EQ(buffer[0], 0xFF);  // Untouched
+}
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -21,6 +21,8 @@
 
 #include "game_lifecycle.h"
 #include "integration_test_hooks.h"
+#include "context.h"
+#include "entrance.h"
 #include <ship/resource/ResourceManager.h>
 #include <ship/resource/ResourceLoader.h>
 #include <ship/resource/archive/ArchiveManager.h>
@@ -360,6 +362,29 @@ void MM_Game_Suspend(void) {
 void MM_Game_Resume(void) {
     fprintf(stderr, "[MM] Game_Resume called\n");
     fflush(stderr);
+
+    // Restore the frozen MM SaveContext captured before we left for OoT (#170).
+    // OoT may have scribbled over the unified gSaveContext storage while it
+    // was active (see src/common/unified_save.c), so we must re-hydrate MM's
+    // view on every resume. First boot of MM has no frozen state — in that
+    // case MM_InitFirstEntrySaveContext still handles bootstrap via #168.
+    if (Context_HasFrozenState(GAME_MM)) {
+        fprintf(stderr, "[MM] Restoring frozen SaveContext on resume\n");
+        fflush(stderr);
+        Context_RestoreState(GAME_MM, &gSaveContext, sizeof(gSaveContext));
+
+        // Prefer the startup entrance set for this switch; fall back to the
+        // return entrance recorded when we last froze MM.
+        uint16_t startup = Combo_GetStartupEntrance();
+        uint16_t returnEntrance = Context_GetFrozenReturnEntrance(GAME_MM);
+        uint16_t targetEntrance = startup != 0 ? startup : returnEntrance;
+        if (targetEntrance != 0) {
+            gSaveContext.save.entrance = targetEntrance;
+            fprintf(stderr, "[MM] Resume entrance: 0x%04X (startup=%u)\n",
+                    targetEntrance, startup != 0);
+        }
+    }
+
     // TODO: Restore MM audio, reload MM-specific resources if needed
 }
 
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -374,15 +374,16 @@ void MM_Game_Resume(void) {
         Context_RestoreState(GAME_MM, &gSaveContext, sizeof(gSaveContext));
 
         // Prefer the startup entrance set for this switch; fall back to the
-        // return entrance recorded when we last froze MM.
-        uint16_t startup = Combo_GetStartupEntrance();
-        uint16_t returnEntrance = Context_GetFrozenReturnEntrance(GAME_MM);
-        uint16_t targetEntrance = startup != 0 ? startup : returnEntrance;
-        if (targetEntrance != 0) {
-            gSaveContext.save.entrance = targetEntrance;
-            fprintf(stderr, "[MM] Resume entrance: 0x%04X (startup=%u)\n",
-                    targetEntrance, startup != 0);
-        }
+        // return entrance recorded when we last froze MM. Use the explicit
+        // Has accessor because entrance 0x0000 is a valid id and treating it
+        // as "unset" would silently drop a legitimate restore.
+        bool hasStartup = Combo_HasStartupEntrance();
+        uint16_t targetEntrance = hasStartup
+            ? Combo_GetStartupEntrance()
+            : Context_GetFrozenReturnEntrance(GAME_MM);
+        gSaveContext.save.entrance = targetEntrance;
+        fprintf(stderr, "[MM] Resume entrance: 0x%04X (startup=%u)\n",
+                targetEntrance, hasStartup);
     }
 
     // TODO: Restore MM audio, reload MM-specific resources if needed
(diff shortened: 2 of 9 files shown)


spencerduncan:

spencerduncan:
## Review: Request changes — design is right, build is broken

**Verdict:** 🛑 Request changes. The architecture, dispatch fix, restore logic, re-entrancy guard, and test suite are all sound and solve the diagnosed bug cleanly. But the patch does not compile — CI has confirmed build failures on Linux and Windows. See specifics below.

### ❌ Build failure (blocking)

Both `build-linux` and `build-windows` fail at `games/oot/soh/GameExports_SingleExe.cpp`:

```
GameExports_SingleExe.cpp:176: error: 'gSaveContext' was not declared in this scope
GameExports_SingleExe.cpp:184: error: 'gSaveContext' was not declared in this scope
```

`integration-tests-linux` fails for the same reason (same TU compiled by that job).

**Cause.** The PR adds `Context_RestoreState(GAME_OOT, &gSaveContext, …)` and `gSaveContext.entranceIndex = targetEntrance` inside `OoT_Game_Resume` (near line 176–184 post-patch), but `extern \"C\" SaveContext gSaveContext;` in this TU is declared way down at line ~237, just above `Combo_CheckEntranceSwitch` where it was previously the only use site. The new use in `Game_Resume` is above the declaration, so the compiler rejects it.

**Fix.** Move `extern \"C\" SaveContext gSaveContext;` up into the top-of-file `extern \"C\"` block (alongside `gAudioContextInitalized` at ~line 32), or duplicate the declaration just before `OoT_Game_Resume`. The MM file doesn't hit this because it pulls in `global.h`, which declares `gSaveContext` for the MM layout.

### ✅ Dispatch fix — correct

`Combo_CheckEntranceSwitch` now resolves the source game via `Context_GetCurrentGame()` and calls `Combo_CheckCrossGameEntrance(gameId, …)` against the right entrance table. The early-boot `\"oot\"` default is safe given OoT always boots first, and `main.cpp` now keeps `Context_SetCurrentGame` in sync after `GameRunner_StartGame` and after every successful `GameRunner_SwitchTo`. Good.

One minor caveat worth calling out (the inline comment already does, but flagging for reviewers): when MM invokes this shared symbol, `sizeof(gSaveContext)` in the OoT TU still resolves to OoT's struct size. That's benign because `Context_FreezeState` clamps to the per-game N64 size internally, and the unified buffer is byte-addressed identically on both sides — but it means the `size` arg passed is effectively a lower bound, not ground truth. Fine as-is.

### ✅ SaveContext restore logic — correct

- `OoT_Game_Resume` restores only when `Context_HasFrozenState(GAME_OOT)` is true — first boot correctly skips.
- Startup-entrance preferred over frozen return-entrance matches how `Play_Init` consumes `Combo_GetStartupEntrance` on the next scene init; frozen fallback covers the mid-session resume case.
- Mirror fix in `MM_Game_Resume` writes to `gSaveContext.save.entrance` (MM's struct layout), which is the right field name for MM's `SaveContext`. Good.

### ✅ Re-entrancy guard — correct

`if (Combo_IsCrossGameSwitch()) return entranceIndex;` at the top of `Combo_CheckEntranceSwitch` prevents a second entrance-hook call on a subsequent transition frame from re-freezing on top of the already-captured state. This is the right place for the guard — earlier than the `Combo_CheckCrossGameEntrance` call that could otherwise mutate pending-switch state.

### ✅ Tests — solid, one small nit

The 4 new gtests in `ReturnPathTest` are a targeted regression suite for #170:
- `MMSideCheckFindsReturnEntrance` — covers the leg that silently no-op'd pre-fix.
- `OoTSideCheckDoesNotMatchMMEntrance` — negative guard against the old hardcoded-\"oot\" behavior.
- `FullRoundtripPreservesOoTSaveContext` — fingerprints both SaveContexts and asserts coexistence.
- `RestoreWithoutFreezeIsSafe` — edge case where `OoT_Game_Resume` reached without a prior freeze.

The `Test_Roundtrip` CTest now exercises `Combo_*` (the production C API) instead of the internal `Entrance_*` C++ API, which is the correct layer to lock in this regression.

**Nit:** The new tests hit `Combo_CheckCrossGameEntrance` directly rather than `Combo_CheckEntranceSwitch` (the fixed dispatch wrapper). That's reasonable — the dispatch logic is thin — but it means the `Context_GetCurrentGame()` branch in `GameExports_SingleExe.cpp` has no direct unit coverage; only the CTest `Roundtrip` indirectly validates it end-to-end. Fine for now, but worth a follow-up to parameterize the dispatch wrapper over `(currentGame, entranceId) → expectedTarget` if this code grows.

### Summary

Fix the `gSaveContext` declaration order and this is ready to ship. Design, restore logic, re-entrancy guard, and regression tests are all well-targeted for the root cause described.

spencerduncan (commented):

# Code review — PR #240 (OoT state restoration on return from MM, #170)

**Recommendation:** REQUEST CHANGES (1 blocking, 3 should-fix, 3 nits)

Files touched:
- `games/oot/soh/GameExports_SingleExe.cpp`
- `games/mm/2s2h/GameExports_SingleExe.cpp`
- `rsbs/src/main.cpp`
- `src/common/test_runner.cpp`
- `combo/tests/context_test.cpp`

Overall the approach is sound: add an explicit `extern` for `gSaveContext` at the top of the OoT TU (fixing the prior ordering bug), re-hydrate the per-game SaveContext on every `*_Game_Resume` from the frozen store, track the active game in `Context` so `Combo_CheckEntranceSwitch` dispatches to the right entrance table, and beef up tests to exercise both legs of the round-trip via the production C API. The OoT changes are well-commented and match existing file conventions. The MM-side changes and a couple of small correctness issues keep this from being an approve.

---

## Blocking (1)

### B1. MM `GameExports_SingleExe.cpp` — no explicit `extern` for `gSaveContext`; relies on a transitive include

**File:** `games/mm/2s2h/GameExports_SingleExe.cpp` (new body of `MM_Game_Resume`)

The MM TU newly dereferences `gSaveContext` (for `Context_RestoreState(GAME_MM, &gSaveContext, sizeof(gSaveContext))` and `gSaveContext.save.entrance = targetEntrance;`) but, unlike the OoT TU in this same PR, does **not** add an explicit `extern SaveContext gSaveContext;` inside the file. The declaration must be coming in transitively through `global.h` (or nothing — in which case this won't compile, matching the failing CI).

Two problems with this:

1. **Symmetry / maintainability.** The OoT side of this PR explicitly does:
   ```cpp
   // OoT's SaveContext (type defined via OTRGlobals.h -> z64save.h).
   // Declared here so OoT_Game_Resume() can restore it on return from MM (#170).
   extern SaveContext gSaveContext;
   ```
   …exactly because the previous incarnation hit a use-before-extern bug. Leaving the MM side to fish the declaration out of `global.h` re-creates the same fragility the OoT side was just fixed against — and the MM file's other externs (`MM_gScreenWidth`, `MM_gSystemHeap`, `MM_gPadMgr`, …) are all explicitly declared in its `extern "C" { … }` block for precisely this reason.
2. **Likely root cause of the failing CI.** If `global.h` doesn't resolve to a TU that declares `SaveContext gSaveContext;` (or declares it with MM's layout), you'll either see `'gSaveContext' was not declared in this scope` or a type mismatch / incomplete-type error on `sizeof(gSaveContext)`.

**Fix:** mirror the OoT pattern. Inside the existing `extern "C" { … }` block near the top of the MM file, add:

```cpp
// MM's SaveContext (type defined via 2s2h global.h -> z64save.h).
// Declared here so MM_Game_Resume() can restore it on return from OoT (#170).
extern SaveContext gSaveContext;
```

If `SaveContext` isn't already visible at that point, also `#include "z64save.h"` (or whichever header the MM tree exposes its layout via) alongside the new `context.h` / `entrance.h` includes.

---

## Should-fix (3)

### S1. `0` is a valid entrance index — don't use it as a "no startup entrance" sentinel

**Files:** both `GameExports_SingleExe.cpp` files, inside the resume bodies.

```cpp
uint16_t startup       = Combo_GetStartupEntrance();
uint16_t returnEntrance = Context_GetFrozenReturnEntrance(GAME_OOT); // or GAME_MM
uint16_t targetEntrance = startup != 0 ? startup : returnEntrance;
if (targetEntrance != 0) {
    gSaveContext.entranceIndex = targetEntrance;         // OoT
    // gSaveContext.save.entrance = targetEntrance;      // MM
    …
}
```

Entrance index `0x0000` is a real, valid entrance in both games (e.g. OoT `ENTR_KOKIRI_FOREST_0` is literally index 0). Treating `0` as "no entrance set" means:

- A legitimate restore to entrance `0` is silently skipped (both the `startup != 0` check and the outer `targetEntrance != 0` check will drop it).
- If the frozen return entrance is `0`, the fallback also fails and `gSaveContext.entranceIndex` is left as whatever MM (or OoT) left it — i.e. the "unified storage got scribbled on" case that this whole PR exists to fix.

**Fix:** use a dedicated sentinel. Either:

- Change `Combo_GetStartupEntrance()` / `Context_GetFrozenReturnEntrance()` to return `ENTR_LOAD_OPENING` / `0xFFFF` / `(uint16_t)-1` as the "unset" value, or
- Pair them with an `bool Combo_HasStartupEntrance()` / `bool Context_HasFrozenReturnEntrance(GameId)` probe and gate the assignment on that.

Same bug exists on both OoT and MM sides, introduced identically.

### S2. The new early-return in `Combo_CheckEntranceSwitch` changes the return value on mid-transition frames

**File:** `games/oot/soh/GameExports_SingleExe.cpp`, `Combo_CheckEntranceSwitch`.

```cpp
extern "C" uint16_t Combo_CheckEntranceSwitch(uint16_t entranceIndex) {
    if (Combo_IsCrossGameSwitch()) {
        return entranceIndex;                 // <-- new
    }
    …
    uint16_t result = Combo_CheckCrossGameEntrance(gameId, entranceIndex);
    …
    return result;                            // (old path)
}
```

Pre-PR, this function unconditionally ran `Combo_CheckCrossGameEntrance` and returned whatever it produced, so even on a "switch is already queued" frame the entrance index caller got could have been an overridden value. Now, if a switch is queued, the caller gets back the **unmodified** `entranceIndex` — which means `randomizer_entrance.c`'s `Entrance_OverrideNextIndex` will feed the raw pre-override entrance into OoT's transition logic for one or more frames until the main loop consumes the pending switch.

In practice this is probably fine (the frame never actually plays out because the switch happens first), but:

1. It's a behavioral change the commit message doesn't mention — worth a sentence in the doxygen comment next to it, spelling out why `entranceIndex` (not the possibly-overridden result) is what we want to return here.
2. Confirm that no caller treats the return of `Combo_CheckEntranceSwitch` as "here's the entrance you should transition to"; if one does, it'll follow a stale link for a frame.

### S3. MM and OoT trust `sizeof(gSaveContext)` to mean "this TU's SaveContext layout", but the unified storage is a single buffer

**Files:** both `GameExports_SingleExe.cpp`.

On OoT:
```cpp
Context_RestoreState(GAME_OOT, &gSaveContext, sizeof(gSaveContext));
```
On MM:
```cpp
Context_RestoreState(GAME_MM, &gSaveContext, sizeof(gSaveContext));
```

The comment in `Combo_CheckEntranceSwitch` already acknowledges this nuance for freezes:

> sizeof(gSaveContext) in this TU is OoT's SaveContext layout, but the underlying unified storage is identical regardless of caller and Context_FreezeState clamps to the per-game N64 size anyway.

Two asks:

- Duplicate that reassurance (or a short version of it) next to the restore calls too — otherwise a future reader sees `sizeof(gSaveContext)` on the MM side and worries we're memcpying MM's (larger) struct over OoT's layout.
- Please verify in the code that `Context_RestoreState` really does clamp to `min(storedSize, OOT_SAVE_CONTEXT_SIZE / MM_SAVE_CONTEXT_SIZE)` rather than to the `size` arg. If it clamps to `size`, MM side would try to write MM's full sizeof over the unified buffer, which may still be safe but is worth an explicit assertion.

---

## Nits (3)

### N1. `fprintf(stderr, "[OoT] Resume entrance: 0x%04X (startup=%u)\n", targetEntrance, startup != 0);`

You're printing `startup != 0` (a bool coerced to `0` or `1`) under `%u` and labeling it `startup=` — that's confusing. Either log the actual value (`startup`) so the log shows both the raw startup and the resolved target, or rename the label to `startupPresent=` / `hadStartup=`. Same nit on the MM side.

### N2. Duplicate `Context_ClearAllFrozenStates()` at start of `Test_Roundtrip`

```cpp
Context_InitFrozenStates();
Context_ClearAllFrozenStates();   // added
```

If `Context_InitFrozenStates` doesn't reset state on re-entry, good; if it does, the clear is redundant. Small thing, but worth a one-line comment explaining which it is ("ensure clean slate in case previous test leaked").

### N3. Long `// =========…` banner comments in `context_test.cpp`

The `// ======` separators in `combo/tests/context_test.cpp` are fine stylistically, but the existing tests in that file (everything above line 420) seem to just use a blank line between tests. Either trim the banners or add them to the earlier tests too — don't love the inconsistency.

---

## Looks good

- The root-cause fix (making `Combo_CheckEntranceSwitch` dispatch on `Context_GetCurrentGame()` instead of hardcoding `"oot"`) is the right call and the doxygen block on the function clearly explains why the single-exe merge makes this necessary.
- Calling `Context_SetCurrentGame` in both the MM-first path and the post-switch path in `rsbs/src/main.cpp` closes the loop so the dispatch actually has something to read.
- Regression tests in `combo/tests/context_test.cpp` are well-scoped — the `OoTSideCheckDoesNotMatchMMEntrance` negative-behavior test is the right guard against the pre-fix bug silently re-appearing, and `RestoreWithoutFreezeIsSafe` is exactly the first-boot edge case I'd want asserted.
- Moving the OoT `extern SaveContext gSaveContext;` to the top of its TU fixes the previously-reported ordering bug cleanly; the old mid-file `extern "C" SaveContext gSaveContext;` is correctly removed in the same diff so there's no duplicate decl.

---

## Namespacing audit (per fork conventions)

Grepped the diff for bare OoT globals that should be `OoT_`-prefixed. Found none of the danger list (`gPlayState`, `Matrix_*`, `Actor_*`, `Flags_*`) — only `gSaveContext`, which is the symbol both TUs' existing code already references bare (OoT's soh and MM's 2s2h layout both land in a single unified buffer in the single-exe build, so this is intentional). No violations to flag.

The MM side does NOT prefix `gSaveContext` as `MM_gSaveContext` even though sibling globals (`MM_gScreenWidth`, `MM_gSystemHeap`, `MM_gPadMgr`, `MM_gSchedContext`) are prefixed — that's consistent with how OoT's file leaves `gSaveContext` bare and matches the "shared unified storage" model, so I wouldn't block on it, but it's worth a design note somewhere (maybe in `unified_save.c`) that `gSaveContext` is deliberately NOT renamed in either tree because it's the shared storage handle.

---

## Summary

- Blocking: 1 (MM file needs the same explicit `extern SaveContext gSaveContext;` the OoT file now has; most likely also the CI failure)
- Should-fix: 3 (0-is-valid-entrance sentinel bug, undocumented early-return semantics change, missing `sizeof` reassurance next to the restore call)
- Nits: 3 (stderr format, duplicate clear, comment-banner style)
- Verdict: **REQUEST CHANGES** — once B1 is fixed and S1 is addressed (S1 is a real correctness bug for entrance 0), this is good to land.

Issue #238, opened 2026-04-20
[Upstream] Sync 2S2H (MM) feature and rando updates


spencerduncan:
## Summary
Sync ~50 safe commits from upstream 2S2H `develop` into `games/mm/`. Note: MM is **not** a subrepo, so this is a manual port.

## Categories
- Check/Item Tracker improvements
- Rando logic fixes (Stone Tower, Snowhead, etc.)
- Menu/tooltip tweaks
- Triforce Hunt display improvements
- Rando option additions
- Minor cosmetic fixes

## Files affected
- `games/mm/2s2h/` (Enhancements, SohGui, Rando)
- `games/mm/mm/src/` (as needed for hooks)

## Dependencies
- Save buffer (#203) — already merged via #223 ✓
- MM AudioLoad (#157) — already merged via #217 ✓

No other hard dependencies.

## Complexity
**L** (Large — ~50 commits across MM subsystems)

## Scoping note
If this proves too large for a single PR, split into:
- Tracker improvements
- Rando logic + options
- UI/cosmetic

## Test criteria
- [ ] MM game boots
- [ ] MM saves load without corruption
- [ ] Check/Item Tracker opens and reflects state
- [ ] New rando options appear and generate valid seeds
- [ ] Triforce Hunt mode works end-to-end
- [ ] No regression in existing MM features
- [ ] Cross-game switching still works (OoT ↔ MM)

## Context
Part of epic #202. Supersedes closed issue #210.

spencerduncan:
@spencerduncan dispatch-agent: task complete, ready for review.

Opened [#242](https://github.com/spencerduncan/redshipblueship/pull/242) — **cosmetic editor trail colors** ([upstream 7078c9bee8](https://github.com/HarbourMasters/2ship2harkinian/commit/7078c9bee8)), a single self-contained low-risk commit.

### Scope note
Upstream 2S2H `develop` is **~488 commits** ahead of `4.0.2` (vs. the issue's ~50 estimate). Per the split-by-category guidance in this issue, I shipped one clean commit first and deferred the rest. Suggested follow-up PRs:

| Category | Notes |
|---|---|
| Enhancements / QoL (~11) | Mostly self-contained — extra powder kegs (#1600), unsheathe sword (#1505), alt toggle animation (#1623/#1624), chateau persistence (#1619), etc. |
[...]

PR #242, opened 2026-04-20
Sync 2S2H (MM): cosmetic editor trail colors (#238, part 1)

spencerduncan:
## Summary
- Ports upstream 2S2H [PR #1491](https://github.com/HarbourMasters/2ship2harkinian/pull/1491) (`7078c9bee8`) — **cosmetic editor trail colors** — to RSBS MM.
- Adds 10 new `Trails` category entries to the MM cosmetic editor covering all swords (Kokiri, Razor, Gilded, Great Fairy, Fierce Deity), Deku stick, Deku spin, Zora punch/kick, and Zora boomerang.
- Wires the new elements into `z_eff_blure.c` (per-weapon trail recolor driven by player form + held item) and `z_en_boom.c` (Zora boomerang trail init).

## Scope
Part of #238 (sync MM with upstream `develop`, 4.0.2..HEAD is ~488 commits). Upstream is far larger than the issue's initial estimate, and several planned categories (BenGui restructuring, save schema, rando) need careful adaptation to the RSBS unified menu and combo save layer. Per the issue's split-by-category guidance, this PR intentionally ships a **single, self-contained, low-risk** commit (cosmetic-only, no game logic, no save layout, no menu restructuring, no new OTR assets). Follow-up PRs will handle enhancements, rando, and infra categories.

## Namespacing
- `gPlayState` → `MM_gPlayState`
- `Effect_GetByIndex` → `MM_Effect_GetByIndex`
- `Actor_ProcessInitChain` call path already uses `MM_Actor_ProcessInitChain` (unchanged).
- `CosmeticEditor_*` / `COSMETIC_ELEMENT_*` live under `games/mm/2s2h/` and follow the existing (unprefixed) pattern.

## Files changed
- `games/mm/2s2h/BenGui/CosmeticEditor.h` — 10 new enum entries inserted between `LIGHT_ARROW_SECONDARY` and `B_BUTTON` (stable ordering, appended within the existing block).
- `games/mm/2s2h/BenGui/CosmeticEditor.cpp` — 10 matching `COSMETIC_EDITOR_ELEMENT` entries with default RGBA and cvar paths.
- `games/mm/src/code/z_eff_blure.c` — `MM_EffectBlure_GetComputedValues` now resolves a trail `elementId` from the player's form / held item when the blure belongs to `player->meleeWeaponEffectIndex[]`, and lerps from the (possibly cosmetic-overridden) start/end colors.
- `games/mm/src/overlays/actors/ovl_En_Boom/z_en_boom.c` — Zora boomerang init reads `COSMETIC_ELEMENT_TRAIL_ZORA_BOOMERANG` for RGB, preserving the original alpha channel.

## Test plan
- [ ] CI: build-windows, build-linux, build-macos, clang-format all pass
- [ ] MM boots to title/file select (smoke)
- [ ] Cosmetic Editor opens, `Trails` category appears with 10 entries, defaults render correctly
- [ ] Each sword form (Kokiri / Razor / Gilded / Great Fairy / Fierce Deity) picks up its own trail color
- [ ] Deku spin and Deku stick (spin attack) pick up their respective trail colors
- [ ] Zora punch vs. Zora kick animations use distinct trail colors
- [ ] Zora boomerang picks up the configured boomerang trail color
- [ ] Disabling (unchecking) a trail in the editor reverts to the stock color
- [ ] No regression in OoT trails (they share no code path; defensive check only)

Refs #238

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634712813.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634588048.zip)


spencerduncan (commented):

# Code Review — PR #242: Sync 2S2H (MM) cosmetic editor trail colors (part 1)

**Branch:** `claude/2s2h-sync-238` @ `30a684c`
**Chrome is not signed into GitHub** — paste this directly into the review form.

**Recommendation: REQUEST CHANGES** — mostly to confirm upstream 2S2H parity on `z_en_boom.c` and split out the unrelated `.claude/dispatch/*.md` churn. If you'd rather ship and follow up, this also reads cleanly as an APPROVE-with-comments since CI is green and the concerns are correctness-adjacent rather than build-breaking.

---

## Summary of diff scope

10 files touched:

- 5 × `.claude/dispatch/*.md` (~83 KB of project-management docs — unrelated to trail colors; see nit #5)
- `games/mm/2s2h/BenGui/CosmeticEditor.h` — adds 10 `COSMETIC_ELEMENT_TRAIL_*` enumerators
- `games/mm/2s2h/BenGui/CosmeticEditor.cpp` — adds 10 matching `COSMETIC_EDITOR_ELEMENT` rows ("Trails" category)
- `games/mm/src/code/z_eff_blure.c` — adds the per-frame element-id resolution + cosmetic override in `MM_EffectBlure_GetComputedValues`
- `games/mm/src/overlays/actors/ovl_En_Boom/z_en_boom.c` — applies cosmetic once in `MM_EnBoom_Init` for the Zora fin boomerang
- `src/common/mm_stubs.c` — passthrough stub for `CosmeticEditor_GetChangedColor` used in single-exe builds

Namespacing check: verified against the tree at `30a684c` — `PLAYER_FORM_FIERCE_DEITY`, `PLAYER_FORM_ZORA`, `PLAYER_FORM_DEKU`, `PLAYER_IA_SWORD_KOKIRI/RAZOR/GILDED/TWO_HANDED`, `PLAYER_IA_DEKU_STICK`, `PLAYER_MWA_ZORA_PUNCH_KICK/JUMPKICK_START/JUMPKICK_FINISH`, `MM_Effect_GetByIndex`, `MM_gPlayState`, `GET_PLAYER`, `COSMETIC_ELEMENT_*`, `CosmeticEditor_GetChangedColor` — all resolve in the renamed MM tree (confirmed `PLAYER_IA_SWORD_TWO_HANDED` exists in `games/mm/src/code/z_player_lib.c`). No `OoT_`-prefixed calls leaked into MM code. ✅

Header signature check: `games/mm/2s2h/BenGui/CosmeticEditor.h` line 31 declares
```c
Color_RGBA8 CosmeticEditor_GetChangedColor(u8 r, u8 g, u8 b, u8 a, u8 elementId);
```
inside an `extern "C"` block. All call sites match this 5-arg form. ✅

---

## Blocking issues (0)

CI is green (build-linux, integration-tests-linux). No symbol-resolution or compile concerns survive after the `mm_stubs.c` passthrough fix.

---

## Should-fix (3)

### 1. Stub return type is a distinct struct from the one declared in the header

`mm_stubs.c`:
```c
typedef struct { uint8_t r, g, b, a; } MM_Stub_Color_RGBA8;
MM_Stub_Color_RGBA8 CosmeticEditor_GetChangedColor(uint8_t r, uint8_t g, uint8_t b, uint8_t a, uint8_t elementId) { ... }
```

`CosmeticEditor.h`:
```c
Color_RGBA8 CosmeticEditor_GetChangedColor(u8 r, u8 g, u8 b, u8 a, u8 elementId);
```

The definition and declaration return **different struct types** (C 6.2.7 compatible-types — same layout but distinct tags → not compatible). This happens to work today because both are 4-byte POD returned in a register on every target you care about, but it's technically UB and will silently break if any target changes the small-struct return ABI, or if someone adds padding to `Color_RGBA8`.

Since the stubs file deliberately avoids project headers (only pulls `<stdint.h>` etc.), the minimal fix is one of:

- Forward-declare with a matching tag: `typedef struct Color_RGBA8_s { uint8_t r, g, b, a; } Color_RGBA8;` — needs to match however `Color_RGBA8` is declared upstream.
- Or return `uint32_t` (packed RGBA) and adjust callers — invasive, probably not worth it.
- Or cleanest: include the canonical color header conditionally.

At minimum, add a comment flagging the ABI assumption so a future refactor doesn't silently break it.

### 2. Enum insertion reorders values — verify preset persistence is by name, not index

The 10 new `COSMETIC_ELEMENT_TRAIL_*` entries are inserted **before** `COSMETIC_ELEMENT_B_BUTTON` in `CosmeticEditor.h`, shifting every subsequent enumerator's numeric value by +10. If any preset/cvar/settings path serializes the element by integer ID (rather than by the cvar string like `"Buttons.B"`), existing user cosmetic configs will map to the wrong elements after upgrade.

Quick scan of `CosmeticEditor.cpp` shows each row has a cvar-path string (4th macro arg, e.g. `"Effects.IceArrowSec"`), which implies the persistence is name-based — but please confirm. If so, this is fine; if any code path round-trips the numeric enum value, append the new trail entries to the end of the enum instead.

### 3. `z_en_boom.c` unconditionally overrides the stock colors

The init loop:
```c
Color_RGBA8 trailColor = CosmeticEditor_GetChangedColor(255, 255, 100, 255, COSMETIC_ELEMENT_TRAIL_ZORA_BOOMERANG);

for (i = 0; i < 4; i++) {
    sp30.p1StartColor[i] = (i < 3) ? <trailColor.{r,g,b}> : D_808A3068[i];
    sp30.p2StartColor[i] = (i < 3) ? <trailColor.{r,g,b}> : D_808A306C[i];
    sp30.p1EndColor[i]   = (i < 3) ? <trailColor.{r,g,b}> : D_808A3070[i];
    sp30.p2EndColor[i]   = (i < 3) ? <trailColor.{r,g,b}> : D_808A3074[i];
}
```

Two things to confirm against 2S2H upstream:

- **RGB flattening:** all four color arrays get the same RGB channels. If the stock arrays `D_808A3068..D_808A3074` had meaningful RGB variation between p1/p2 and start/end (to give the trail a gradient), this change discards that gradient even when the user hasn't customized anything. Compare with `z_eff_blure.c`, which uses the "did the returned value differ from the input?" check (`*(u32*)&newP1Start != *(u32*)&p1Start`) to **only** override when the cosmetic is actually customized — otherwise the stock gradient is preserved.
- **Init-only application:** this runs in `MM_EnBoom_Init`, so mid-flight cosmetic changes won't take effect until the next throw. Fine if upstream does the same, worth a one-line comment if not obvious.

Easiest parity fix is to mirror the `z_eff_blure.c` pattern (compare default-vs-returned, override only on change), but **please confirm against upstream 2S2H behavior first** — if they also flatten, leave it.

---

## Nits (5)

1. **Magic sentinel `255`** in `z_eff_blure.c` (`u8 elementId = 255;` … `if (elementId != 255)`). `COSMETIC_ELEMENT_MAX` is small today but the enum is growing. Use `(u8)-1` or a named `COSMETIC_ELEMENT_NONE` sentinel to be safer.

2. **Strict aliasing / struct punning** in `z_eff_blure.c`:
   ```c
   if (*(u32*)&newP1Start != *(u32*)&p1Start) { ... }
   ```
   Technically UB (C 6.5 §7). Swap to `memcmp(&newP1Start, &p1Start, sizeof(newP1Start)) != 0` or a field-wise compare — compilers optimize both into the same `cmp`.

3. **`int` vs `s32`** in `for (int i = 0; i < 3; i++)` — rest of the file uses `s32`. Minor style consistency.

4. **Alpha semantics comment** — in `z_eff_blure.c`, when the cosmetic is applied, `p1Start.a` / `p2Start.a` come from the cosmetic (via `GetChangedColor`) but `p1End.a` / `p2End.a` are left from the original config (only `.r/.g/.b` are copied from start). That's probably intentional (preserve the original alpha-fade) but worth a one-liner comment so the next reader doesn't "fix" it.

5. **Unrelated `.claude/dispatch/*.md` churn.** 5 files, ~83 KB:
   - `orphan-branch-audit.md`
   - `project-assessment.md`
   - `status-20260419.md`
   - `testing-strategy.md`
   - `worker-prompt-review.md`

   These are project-management / audit docs and have nothing to do with cosmetic trail colors. They balloon the diff, make "Part 1" harder to review, and will clutter `git blame` on unrelated files. Strongly prefer splitting them into a separate PR (or a dedicated `chore: internal docs` branch).

---

## Correctness spot-checks

- **`MM_Effect_GetByIndex` cast pattern:**
  ```c
  this == (EffectBlure*)MM_Effect_GetByIndex(player->meleeWeaponEffectIndex[i])
  ```
  Fine — walks the 3-slot melee-weapon-effect-index array on the player, matches the current `EffectBlure*`. The cast is correct since the melee-weapon effect is always an `EffectBlure` in MM.

- **`PLAYER_MWA_ZORA_PUNCH_KICK` classified as a kick:** the attack-name suggests "punch→kick transition" (3rd attack of the Zora combo ends in a kick). Mapping it into the kick branch alongside `JUMPKICK_START/FINISH` is correct. Worth a sanity-check against upstream 2S2H but matches their logic as described.

- **Fierce Deity vs "Great Fairy" label:** `PLAYER_IA_SWORD_TWO_HANDED` (Fierce Deity's sword when somehow held by Human Link) mapping to `COSMETIC_ELEMENT_TRAIL_SWORD_GREAT_FAIRY` preserves the upstream 2S2H UI label. "Great Fairy Sword" is a slight misnomer in MM canon (there's no such sword in MM) but matches the 2S2H upstream — don't change here, let upstream own the rename.

- **Zora Boomerang coverage split:** the enum's `TRAIL_ZORA_BOOMERANG` is handled only in `z_en_boom.c` (actor), not in the `z_eff_blure.c` player-melee loop — that's correct, the fin boomerang is an `En_Boom` actor, not a player melee weapon.

---

## Part-1 scope / dangling refs

No dangling references found — every new `COSMETIC_ELEMENT_TRAIL_*` enumerator is consumed either in `z_eff_blure.c` (player melee trails) or `z_en_boom.c` (Zora fin boomerang). No call-site references a not-yet-ported symbol. Part 2 should be clean to land on top.

One scope question for the PR description: if #238 has additional trail types queued for part 2 (e.g. spin-attack charge-up trails, hookshot chain, arrow trails with secondary pass), consider listing them in the PR body so reviewers can pre-check for placeholder enum gaps.

---

## Final tally

- **Blocking: 0**
- **Should-fix: 3** (stub ABI correctness, enum-value stability, `z_en_boom.c` unconditional override)
- **Nits: 5**
- **Recommendation: REQUEST CHANGES** — primarily to (a) confirm `z_en_boom.c` matches 2S2H upstream behavior and (b) split the `.claude/dispatch/*.md` docs into a separate PR. If those two are addressed (or justified as intentional), happy to flip to APPROVE.

Issue #233, opened 2026-04-20
[Upstream] Port segfault-on-quit fix (SOH #6212)


spencerduncan:
## Summary
Port upstream SOH commit `79d6f54be` ([#6212](https://github.com/HarbourMasters/Shipwright/pull/6212)) — fix segfault when quitting the game. The fix assigns to file-scope `sohFast3dWindow` (introduced by SOH #5892).

Originally tracked as #206 (closed procedurally).

## Files affected
- `games/oot/soh/OTRGlobals.cpp` (quit teardown)

## Dependencies
**Blocked by:** ImGui-driven extraction flow port (SOH #5892)

## Complexity
**S** (Small — single-file focused fix once `sohFast3dWindow` exists)

## Test criteria
- [ ] OoT quits cleanly (menu → Quit)
- [ ] OoT quits cleanly (window close button)
- [ ] MM quits cleanly (both paths)
- [ ] No segfault reported by OS / Valgrind / sanitizers
- [ ] Cross-game switching does not leak state that then segfaults on quit

## Context
Part of epic #202. Deferred from PR #226 Phase 1 of #211. Supersedes closed issue #206.

spencerduncan:
Merged via PR #301 → main `42e2f5dd163de269bc9a43175edeb40a93128bfb` (squash). Port of SOH `79d6f54be` (#6212), verified byte-identical to the upstream diff: `sohFast3dWindow = nullptr;` after `SohGui::Destroy();` in `DeinitOTR()`.

CI evidence: clang-format, clang-tidy, build-linux, build-windows, integration-tests-linux (BootOoT/BootMM/Roundtrip) all green on PR #301.

Manual test-criteria residue (Valgrind/sanitizers, window-close path, cross-game-switch-then-quit feel) is deferred to the #212 verification carve-out per the lane plan.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #231, opened 2026-04-20
[Upstream] Port portArchiveVersion fix (SOH #6168)


spencerduncan:
## Summary
Port upstream SOH commit `35039565d` ([#6168](https://github.com/HarbourMasters/Shipwright/pull/6168)) — portArchiveVersion fix. Requires introducing `portArchivePath` and `sohArchiveVersionMatch` which are not in the tree.

## Files affected
- `games/oot/soh/OTRGlobals.cpp`
- Archive path/version utility files

## Dependencies
None.

## Complexity
**S** (Small — focused bug fix, but requires two new helpers)

## Test criteria
- [ ] `soh.o2r` version check passes on matching archive
- [ ] Mismatched archive version shows appropriate error
- [ ] Missing archive path handled gracefully

## Context
Part of epic #202. Deferred from PR #226 Phase 1 of #211.

spencerduncan:
Heads-up from delegation triage: the portArchivePath / sohArchiveVersionMatch helpers this issue calls for already exist on main (OTRGlobals.cpp ~lines 289-290 and 344-348), folded in via the #5892/#250 extraction-flow port. Remaining work is only to confirm the version-match + missing-path-graceful behavior matches SOH #6168 (commit 35039565d), then close. Assigned to the OTRGlobals reconcile stream (#211) as verify-only.

spencerduncan:
Merged via PR #302 → main `fa2bc4e57186db9d29654f382dd8b63aa4d902bd` (squash). Port of SOH `35039565d` (#6168), identical to the upstream diff: `DetectOTRVersion(portArchivePath, false)` → `DetectOTRVersion("soh.o2r", false)` at the version-match block in `games/oot/soh/OTRGlobals.cpp`.

Criteria evidence:
- **Version match passes**: `DetectOTRVersion` (OTRGlobals.cpp:1477) resolves the bare filename itself via `LocateFileAcrossAppDirs(fileName, appShortName)`; passing the pre-resolved path double-resolved and missed the archive.
- **Mismatch → "soh.o2r is outdated"**: `RunExtract`'s `ES_PORT_ARCHIVE` popup selects that title when the file exists but versions differ.
- **Missing file graceful**: returns the `{INT16_MAX, INT16_MAX, INT16_MAX}` sentinel; popup path selects "Missing soh.o2r".

[...]

Issue #202, opened 2026-02-05
[Epic] Sync with upstream SOH 9.1.2 and 2S2H develop


spencerduncan:
## Overview

RSBS is **232 commits behind** upstream SOH (9.1.1 → 9.1.2+205). This epic tracks the sync effort.

## Analysis Summary

- **~180 commits are SAFE** (isolated features, unlikely to conflict)
- **~30 commits need REVIEW** (touch OTRGlobals, SaveManager)
- **~14 SOH commits are CRITICAL** (bug fixes we need)
- **~9 2S2H commits are CRITICAL** (save buffer, respawn persistence)

### Phase 1: Critical Bug Fixes (Sequential)
- [ ] #203 - Sync 2S2H save buffer (64KB→128KB)
- [ ] #204 - Fix v8 saves crash
- [ ] #205 - Fix AppImage ROM extraction
- [ ] #206 - Fix segfault on quit

### Phase 2: Feature Batches (Parallel Prep, Sequential Merge)
- [ ] #207 - Randomizer shuffle features (~15 commits)
- [ ] #208 - Dungeon logic refactors (~12 commits)
- [ ] #209 - Enhancement/UI features (~10 commits)
- [ ] #210 - 2S2H feature sync (~50 commits)

### Phase 3: OTRGlobals Reconciliation
- [ ] #211 - Reconcile 30 upstream commits affecting OTRGlobals.cpp

### Phase 4: Integration Verification
- [ ] #212 - Full cross-game test: OoT→MM→OoT, randomizer works

## Merge Protocol

1. Parallel prep: Agents create branches
2. Sequential merge: One batch at a time
3. Namespace re-application between merges
4. Build + boot verification after each merge

## Dependency Graph
```
Phase 1: #203 → #204 → #205 → #206
              ↓
Phase 2: #207, #208, #209, #210 (parallel prep, sequential merge)
              ↓
Phase 3: #211 (OTRGlobals reconciliation)
              ↓
Phase 4: #212 (Integration verification)
```

spencerduncan:
Dispatch agent online. Ingesting state. Verifying handoff: #216 CI in progress, #219 green → merging now, #214 awaiting docs-bypass trigger.

spencerduncan:
[...]

PR #226, opened 2026-04-20
feat: reconcile OTRGlobals.cpp with upstream (#211)

spencerduncan:
## Summary

Phase 1 of reconciling `games/oot/soh/OTRGlobals.cpp` with upstream HarbourMasters/Shipwright `develop` (issue #211). This PR takes a **conservative approach**: it ports only the subset of upstream commits whose changes can be applied cleanly without pulling in missing infrastructure, so CI stays green and the existing RSBS overlays are preserved exactly.

### Fork baseline

The current local `OTRGlobals.cpp` was identified (by exact content match) as derived from upstream commit `14a6bc1f8` ("Add Roc's Feather to Rando Item Pool", 2026-01-10). All ports below are relative to that baseline.

### RSBS overlays preserved (verified after port)

- `RsbsFeatureEnabled()` runtime feature gate (file scope)
- `#ifdef RSBS_SINGLE_EXECUTABLE` singleton-reuse path in the `OTRGlobals` constructor
- All `RsbsFeatureEnabled("RSBS_DISABLE_*")` guards inside `InitOTR()` and `Graph_ProcessGfxCommands`
- `#ifdef SINGLE_EXECUTABLE_BUILD` block with `OoT_FreezeState` / `OoT_ResumeFromContext` / `Combo_*` hooks

### Commits ported in this PR

| Upstream SHA | PR | Description |
|---|---|---|
| `007373646` | [#5973](https://github.com/HarbourMasters/Shipwright/pull/5973) | Fix entrance value assigned to DMT DC sign (`ENTR_DEATH_MOUNTAIN_TRAIL_OUTSIDE_DODONGOS_CAVERN` → `ENTR_DODONGOS_CAVERN_ENTRANCE`) |
| `c68c8f128` | [#5974](https://github.com/HarbourMasters/Shipwright/pull/5974) | Gate entrance labels on loading-zone signs behind `CVAR_RANDOMIZER_ENHANCEMENT(\"EntrancesOnSigns\")` (default 0, so behavior is unchanged until the UI toggle is ported) |
| `f474a8692` (partial) | [#6152](https://github.com/HarbourMasters/Shipwright/pull/6152) | Remove unused `ovl_En_Dns/z_en_dns.h` include; switch two loop indices (`constCameraStrings` copy, `CheckSaveFile` zero-fill) from signed to `size_t` to match operand types |

### Commits deferred (with rationale)

[...]

PR #250, opened 2026-04-21
feat: port ImGui-driven extraction flow from upstream SOH (#5892)

### LINKED ISSUE #227: [Upstream] Port ImGui-driven extraction flow (SOH #5892)
## Summary
Port upstream SOH commit `704ace8fd` ([#5892](https://github.com/HarbourMasters/Shipwright/pull/5892)) — ImGui-driven extraction flow. This is an 833-line refactor introducing `RunExtract` / `PromptSteps` and the file-scope `sohFast3dWindow`.

This is **foundation work** — several other deferred upstream commits depend on symbols introduced here (see PR #226 deferred list).

## Files affected
- `games/oot/soh/OTRGlobals.cpp` (file-scope `sohFast3dWindow`, `RunExtract`, `PromptSteps`)
- `games/oot/soh/Extractor/` (new ImGui flow)
- Related ImGui window files for the extraction prompt

## Unblocks
- #6215 Linux AppImage fix (and follow-ups #6386, #6412, #6501)
- #6212 segfault on quit
- Scrolling Texture Interpolation stacking changes

## Dependencies
None — foundation.

## Complexity
**L** (Large — 833-line refactor touching OTRGlobals + new ImGui window)

## RSBS preservation
Must preserve:
- `RsbsFeatureEnabled()` guards in `InitOTR()` and `Graph_ProcessGfxCommands`
- `#ifdef RSBS_SINGLE_EXECUTABLE` singleton-reuse path in `OTRGlobals` constructor
- `#ifdef SINGLE_EXECUTABLE_BUILD` freeze/resume block

## Test criteria
- [ ] OoT ROM extraction dialog launches on first run
- [ ] ROM extraction completes and O2R is generated
- [ ] Existing O2R cache is reused on subsequent boots
- [ ] Cross-game switching still works
- [ ] All RSBS feature flags still functional

## Context
Part of epic #202. Follows PR #226 Phase 1 of #211 which deferred this commit pending a dedicated PR.

spencerduncan:
## Summary

Closes #227. Ports upstream SOH commit [`704ace8fd`](https://github.com/HarbourMasters/Shipwright/pull/5892) — ImGui-Driven Extraction Flow and Progress Reporting — while preserving all RSBS overlays documented in [apps#226](https://github.com/spencerduncan/redshipblueship/pull/226). This is foundation work that unblocks linux appimage fix (#6215 → #231), segfault-on-quit (#6212 → #233), and the scrolling-texture-interpolation-stacking follow-ups.

### What changes

- **`OTRGlobals::OTRGlobals`** now performs most window/GUI setup that previously lived in `Initialize()`, so there is a `Fast::Fast3dWindow` ready **before** any ROM extraction runs. This lets the extraction UI be driven from ImGui instead of OS dialogs and SDL message boxes.
- **New `OTRGlobals::RunExtract`** drives a state machine (`ES_PORT_ARCHIVE → ES_WINDOWS → ES_EXTRACT_ARGS → ES_EXTRACT → ES_VERIFY`) with `PromptSteps` / `WindowsSteps` sub-states. Replaces the big inline block of Windows/macOS/Linux/Switch/WiiU branches that used to live inside `InitOTR`.
- **`SohGui::SetupGuiElements`** is split into `SetupMenu` / `SetupMenuElements` / `SetupGuiElements`. `SohMenu::DrawElement` now guards on a `mMenuElementsInitialized` flag so the menu does not attempt to draw before its entries are populated.
- **`SohModals`** gains `PopupsQueued()` and centers modals on the main viewport (fixes the early-startup modals being off-screen). `SohGui` gains matching `PopupsQueued()` / `DismissPopup()` wrappers used by the new state machine.
- **`Extract.h/cpp`** match the upstream `CallZapd` signature with optional `std::atomic<size_t>*` progress pointers. The pinned ZAPDTR submodule still only exposes `zapd_main` (not `zapd_report`), so those pointers currently pass through without being updated and the progress popup shows an indeterminate \"Starting Up\" state during extraction. Wiring up real progress reporting requires an independent ZAPD uplift — intentionally out of scope for this PR.

### RSBS overlays preserved (verified)

- `RsbsFeatureEnabled()` and all 11 `RSBS_DISABLE_*` guards inside `InitOTR` and `Graph_ProcessGfxCommands`.
- `#ifdef RSBS_SINGLE_EXECUTABLE` early-return in the `OTRGlobals` constructor. `RunExtract` also checks `sohFast3dWindow == nullptr` under the same ifdef, so single-exe mode skips the extraction flow entirely — the parent launcher owns that.
- `#ifdef SINGLE_EXECUTABLE_BUILD` freeze/resume block at the bottom of the file.
- `OoT_` prefixed references to `gBuildVersion*` and `gPlayState`.
- `ShipInit::InitAll()` moves to the end of `InitOTR` (matching upstream's move) while keeping the `RSBS_DISABLE_SHIPINIT` guard.

### Deliberate omissions from upstream `704ace8fd`

- `Randomizer::CreateCustomMessages()` is **not** re-added. [apps#226](https://github.com/spencerduncan/redshipblueship/pull/226) noted that target line is already absent locally (removed by upstream `9c1a1728e` before the fork baseline) and deferred any restoration to a follow-up.
- Local `Extract.cpp` keeps its existing `try`/`catch` safety around `zapd_main` but drops the Win32 `ShowConsoleWindow` dance and the linux `MessageboxWorker` thread — the ImGui progress popup replaces them.
- `extern \"C\" void PadMgr_ThreadEntry(PadMgr*);` from upstream is unrelated to this refactor and unused inside the commit; skipped.

## Test plan

- [ ] `build-windows` passes
- [ ] `build-linux` passes
- [ ] `build-macos` passes
- [ ] `clang-format` passes
- [ ] Manual smoke: fresh profile — OoT ROM extraction dialog appears on first run, completes, and the O2R is generated
- [ ] Manual smoke: relaunch on existing O2R cache boots straight into the title screen without re-prompting
- [ ] Manual smoke: cross-game switching still functions after extraction
- [ ] Manual verification: all `RSBS_DISABLE_*` env-var toggles still skip their subsystems (spot-check `RSBS_DISABLE_AUDIO`, `RSBS_DISABLE_SHIPINIT`, `RSBS_DISABLE_RANDO`)
- [ ] Manual verification: single-exe build boots OoT without null-deref on `sohFast3dWindow`

Refs #227. Part of epic #202. Follow-ups on #231, #233 unblocked.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634528647.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6634450142.zip)

### DIFF
--- a/games/oot/soh/Extractor/Extract.cpp
+++ b/games/oot/soh/Extractor/Extract.cpp
@@ -406,7 +406,6 @@ bool Extractor::ManuallySearchForRom() {
     std::ifstream inFile;
 
     if (!GetRomPathFromBox()) {
-        ShowErrorBox("Ocarina of Time - No rom selected", "No Ocarina of Time ROM selected. Exiting");
         return false;
     }
 
@@ -486,11 +485,16 @@ bool Extractor::RunFileStandalone(std::string rom) {
     return true;
 }
 
+void Extractor::SetSearchPath(const std::string& path) {
+    mSearchPath = path;
+}
+
 bool Extractor::Run(std::string searchPath, RomSearchMode searchMode, std::string installPath) {
     std::vector<std::string> roms;
     std::ifstream inFile;
 
-    mSearchPath = searchPath;
+    SetSearchPath(searchPath);
+
     GetRoms(roms);
 
     // On Linux AppImage (and for users who drop a ROM next to the install), also
@@ -648,9 +652,11 @@ std::string Extractor::Mkdtemp() {
 }
 
 extern "C" int zapd_main(int argc, char** argv);
-static void MessageboxWorker();
 
-bool Extractor::CallZapd(std::string installPath, std::string exportdir) {
+bool Extractor::CallZapd(std::string installPath, std::string exportdir, std::atomic<size_t>* extractCount,
+                         std::atomic<size_t>* totalExtract) {
+    (void)extractCount;
+    (void)totalExtract;
     constexpr int argc = 22;
     char xmlPath[1024];
     char confPath[1024];
@@ -723,47 +729,25 @@ bool Extractor::CallZapd(std::string installPath, std::string exportdir) {
         return false;
     }
 
-#ifdef _WIN32
-    // Grab a handle to the command window.
-    HWND cmdWindow = GetConsoleWindow();
-
-    // Normally the command window is hidden. We want the window to be shown here so the user can see the progess of the
-    // extraction.
-    ShowWindow(cmdWindow, SW_SHOW);
-    SetWindowPos(cmdWindow, HWND_NOTOPMOST, 0, 0, 0, 0, SWP_NOSIZE | SWP_NOMOVE);
-#else
-    // Show extraction in background message until linux/mac can have visual progress
-    std::thread mbThread(MessageboxWorker);
-    mbThread.detach();
-#endif
-
+    // NOTE: progress reporting via zapd_report is not yet wired up in this tree; the ZAPD submodule
+    // pinned here still exposes only zapd_main. The ImGui progress popup owned by RunExtract will
+    // show an indeterminate "Starting Up" state until ZAPD is uplifted.
     try {
         zapd_main(argc, (char**)argv.data());
     } catch (const std::exception& e) {
         fprintf(stderr, "Extraction failed: %s\n", e.what());
         std::string errMsg = "ZAPD extraction failed with error: " + std::string(e.what());
         ShowErrorBox("Extraction Failed", errMsg.c_str());
-#ifdef _WIN32
-        ShowWindow(cmdWindow, SW_HIDE);
-#endif
         std::filesystem::current_path(curdir);
         std::filesystem::remove_all(tempdir);
         return false;
     } catch (...) {
         fprintf(stderr, "Extraction failed with unknown error\n");
         ShowErrorBox("Extraction Failed", "ZAPD extraction failed with an unknown error.");
-#ifdef _WIN32
-        ShowWindow(cmdWindow, SW_HIDE);
-#endif
         std::filesystem::current_path(curdir);
         std::filesystem::remove_all(tempdir);
         return false;
     }
-
-#ifdef _WIN32
-    // Hide the command window again.
-    ShowWindow(cmdWindow, SW_HIDE);
-#endif
 
     // Verify the output file was created before attempting to copy
     if (!std::filesystem::exists(otrFile)) {
@@ -782,11 +766,4 @@ bool Extractor::CallZapd(std::string installPath, std::string exportdir) {
     std::filesystem::remove_all(tempdir);
 
     return true;
-}
-
-static void MessageboxWorker() {
-    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_INFORMATION, "Extracting",
-                             "Extraction will now begin in the background.\n\nPlease be patient for the process to "
-                             "finish. Do not close the main program.",
-                             nullptr);
 }
--- a/games/oot/soh/Extractor/Extract.cpp
+++ b/games/oot/soh/Extractor/Extract.cpp
@@ -682,88 +682,98 @@ bool Extractor::CallZapd(std::string installPath, std::string exportdir, std::at
     // Work this out in the temporary folder
     std::string tempdir = Mkdtemp();
     std::string curdir = std::filesystem::current_path().string();
+
+    // CallZapd is invoked from a detached worker thread by the ImGui extraction flow, so any
+    // uncaught exception would std::terminate the process. Wrap the whole body and guarantee
+    // cwd + tempdir cleanup on every exit path (including std::filesystem::copy failures).
+    bool success = false;
+    try {
 #ifdef _WIN32
-    std::filesystem::copy(assetsPath, tempdir + "/assets",
-                          std::filesystem::copy_options::recursive | std::filesystem::copy_options::update_existing);
+        std::filesystem::copy(
+            assetsPath, tempdir + "/assets",
+            std::filesystem::copy_options::recursive | std::filesystem::copy_options::update_existing);
 #else
-    std::filesystem::create_symlink(assetsPath, tempdir + "/assets");
+        std::filesystem::create_symlink(assetsPath, tempdir + "/assets");
 #endif
 
-    std::filesystem::current_path(tempdir);
+        std::filesystem::current_path(tempdir);
 
-    snprintf(xmlPath, 1024, "assets/xml/%s", version);
-    snprintf(confPath, 1024, "assets/Config_%s.xml", version);
-    snprintf(portVersion, 18, "%d.%d.%d", OoT_gBuildVersionMajor, OoT_gBuildVersionMinor, OoT_gBuildVersionPatch);
+        snprintf(xmlPath, 1024, "assets/xml/%s", version);
+        snprintf(confPath, 1024, "assets/Config_%s.xml", version);
+        snprintf(portVersion, 18, "%d.%d.%d", OoT_gBuildVersionMajor, OoT_gBuildVersionMinor, OoT_gBuildVersionPatch);
 
-    argv[0] = "ZAPD";
-    argv[1] = "ed";
-    argv[2] = "-i";
-    argv[3] = xmlPath;
-    argv[4] = "-b";
-    argv[5] = romPath.c_str();
-    argv[6] = "-fl";
-    argv[7] = "assets/filelists";
-    argv[8] = "-gsf";
-    argv[9] = "0";
-    argv[10] = "-rconf";
-    argv[11] = confPath;
-    argv[12] = "-se";
-    argv[13] = "OTR";
-    argv[14] = "--otrfile";
-    argv[15] = otrFile;
-    argv[16] = "--portVer";
-    argv[17] = portVersion;
-    argv[18] = "-o";
-    argv[19] = "placeholder";
-    argv[20] = "-osf";
-    argv[21] = "placeholder";
+        argv[0] = "ZAPD";
+        argv[1] = "ed";
+        argv[2] = "-i";
+        argv[3] = xmlPath;
+        argv[4] = "-b";
+        argv[5] = romPath.c_str();
+        argv[6] = "-fl";
+        argv[7] = "assets/filelists";
+        argv[8] = "-gsf";
+        argv[9] = "0";
+        argv[10] = "-rconf";
+        argv[11] = confPath;
+        argv[12] = "-se";
+        argv[13] = "OTR";
+        argv[14] = "--otrfile";
+        argv[15] = otrFile;
+        argv[16] = "--portVer";
+        argv[17] = portVersion;
+        argv[18] = "-o";
+        argv[19] = "placeholder";
+        argv[20] = "-osf";
+        argv[21] = "placeholder";
 
-    // Validate config XML exists before calling zapd_main
-    if (!std::filesystem::exists(confPath)) {
-        std::string errMsg = "Extractor config not found: " + std::string(confPath) +
-                             "\n\nThis may indicate an incomplete installation.";
-        fprintf(stderr, "Extractor config not found: %s. This may indicate an incomplete installation.\n", confPath);
-        ShowErrorBox("Extraction Failed", errMsg.c_str());
-        std::filesystem::current_path(curdir);
-        std::filesystem::remove_all(tempdir);
-        return false;
-    }
+        // Validate config XML exists before calling zapd_main
+        if (!std::filesystem::exists(confPath)) {
+            std::string errMsg = "Extractor config not found: " + std::string(confPath) +
+                                 "\n\nThis may indicate an incomplete installation.";
+            fprintf(stderr, "Extractor config not found: %s. This may indicate an incomplete installation.\n",
+                    confPath);
+            ShowErrorBox("Extraction Failed", errMsg.c_str());
+        } else {
+            // NOTE: progress reporting via zapd_report is not yet wired up in this tree; the ZAPD
+            // submodule pinned here still exposes only zapd_main. The ImGui progress popup owned by
+            // RunExtract will show an indeterminate "Starting Up" state until ZAPD is uplifted.
+            try {
+                zapd_main(argc, (char**)argv.data());
+            } catch (const std::exception& e) {
+                fprintf(stderr, "Extraction failed: %s\n", e.what());
+                std::string errMsg = "ZAPD extraction failed with error: " + std::string(e.what());
+                ShowErrorBox("Extraction Failed", errMsg.c_str());
+                throw;
+            } catch (...) {
+                fprintf(stderr, "Extraction failed with unknown error\n");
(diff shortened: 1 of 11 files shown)


spencerduncan:

spencerduncan:
# PR #250 Code Review — ImGui Extraction Flow

Read-only review, no changes made. Overall this is a faithful port of upstream 704ace8fd and the RSBS overlays that #226 / #227 called out are preserved. Flagging a few real concerns plus a handful of smaller ones.

## Blocking / needs a second look

### 1. Use-after-scope on `tfile` in `WS_PERMS` lambda (RunExtract)
`games/oot/soh/OTRGlobals.cpp` — `RunExtract` / `WS_PERMS` case:

```cpp
case WS_PERMS: {
    FILE* tfile = fopen("./text.txt", "w");
    ...
    if (tfile == NULL || error) {
        SohGui::RegisterPopup("SoH Permissions Error", ..., "OK", "", [&]() {
            fclose(tfile);
            PathTestCleanup(tfile);
            exit(0);
        });
    }
    ...
    continue;
}
```

`tfile` is a local inside the `case WS_PERMS` block. The lambda captures it by reference. `RegisterPopup` is asynchronous — the popup fires on a later main-loop iteration, by which point the `case` block has been left and the storage for `tfile` is no longer valid. In practice `fclose(tfile)` inside the lambda is a double-jeopardy call anyway (the "good" branch already `fclose`'d it before queuing, and the "error" branch has `tfile == NULL`), but the fundamental capture-by-reference of a block-scoped `FILE*` is UB.

Upstream has the same pattern. Worth fixing here, or explicitly noting why it's safe.

### 2. Silent deletion of existing O2R files on any version mismatch
`games/oot/soh/OTRGlobals.cpp`:

```cpp
bool shouldRegen = VerifyArchiveVersion(vanillaVersion) || VerifyArchiveVersion(mqVersion);
...
} else if (shouldRegen) {
    SohGui::RegisterPopup("Outdated ROM Archives",
                          "Your oot.o2r or oot-mq.o2r were created with incompatible versions of SoH.\nYou will "
                          "now be redirected to re-extract them.");
    std::filesystem::remove("oot.o2r");
    std::filesystem::remove("oot-mq.o2r");
}
```

Two issues:
- The `remove` calls run **immediately**, not from the popup callback. The "informational" popup is registered then deletion happens on the same tick — the user has not confirmed, and they can't say "no." Previously this prompted with Yes/No ("Would you like to regenerate it now?"). That's a UX regression.
- `VerifyArchiveVersion` returns `true` whenever `major != INT16_MAX && major != OoT_gBuildVersionMajor`. If an O2R file exists but the version block reads `{0, 0, 0}` (old/corrupt), we silently delete it. Users who have invested ROM extraction time could lose it on a flaky read.

At minimum: move the `remove`s into the popup callback after user confirmation, or restore the Yes/No prompt.

### 3. Crash handler initialization is now post-extraction
Pre-refactor, `context->InitCrashHandler()` ran before the extraction block. This refactor moves `InitConsole`, `InitControlDeck`, `InitResourceManager`, `InitWindow` into the constructor but **leaves `InitCrashHandler()` in `Initialize()`**. `Initialize()` is still only called after `RunExtract` returns. If ZAPD throws, a filesystem call explodes, or ImGui asserts during the extraction UI, there is no crash handler. This matches upstream 704ace8fd (they have the same issue), but worth a comment acknowledging it or moving the call up.

## Worth checking before merge

### 4. `SohGui::SetupMenu()` is called in the ctor before the `sohArchiveVersionMatch` branch
`SetupMenu` calls `gui->SetMenu(mSohMenu)` and `gui->AddGuiWindow(mModalWindow); mModalWindow->Show();`. This runs *regardless* of whether `soh.o2r` is present or version-matched. If `InitResourceManager({ portArchivePath }, {}, 3, true)` succeeds-with-warnings on a missing `soh.o2r`, the menu/modal objects get created, the font-loading block is skipped, and the state machine's `Missing soh.o2r` popup runs with ImGui's default font. Probably fine, but worth smoke-testing the "soh.o2r missing" path end-to-end: the missing-archive popup must render readably before `exit(1)`.

### 5. `ImGui::EndPopup()` brace fix in `SohModals.cpp`
Correct fix — `EndPopup()` must only run when `BeginPopupModal` returned true. The diff moves `ImGui::EndPopup()` inside the `if` block. Good catch from upstream, this was a latent bug.

### 6. `sohFast3dWindow` promoted to file-scope global
Previously a local in `Initialize()`, now a bare `std::shared_ptr<Fast::Fast3dWindow> sohFast3dWindow;` at file scope (not `static`, no `extern`). Two concerns:
- Non-`static` at namespace scope means external linkage. If any other TU ever defines a symbol with the same name, ODR violation. Should be `static` or in an anonymous namespace.
- `portArchivePath` has the same issue.

Cheap defensive fix: add `static` to both.

### 7. Lambda lifetime of `extract`, `file`, `args` across threadpool
`RunExtract` holds these as locals and passes them into `threadPool->detach_task(...)` lambdas by reference. The `shared_ptr<BS::thread_pool>` is also a local — its destructor (when `RunExtract` returns) must wait for pending tasks, otherwise background tasks access freed stack. `BS::thread_pool`'s dtor does wait by default; confirm the pinned version has that behavior.

Also: `extract.CallZapd` runs on the background thread, and the detached task captures `extract` by reference while the main thread is still iterating the state machine. If the main thread ever reassigns `extract = Extractor();` while a task is in flight, you get a race. Looking at the state machine, each `extract = Extractor()` happens before `extracting = true` is set and no new assignment happens until `extracting` goes back to `false`. Looks safe, but it's subtle — worth a comment.

### 8. `sohArchiveVersionMatch` only set in the ctor
If `RSBS_SINGLE_EXECUTABLE` takes the early return in the ctor, `sohArchiveVersionMatch` remains at its static-init `false`. `RunExtract` has a matching `sohFast3dWindow == nullptr` guard that early-returns, so the stale `false` is never observed. Correct today — but the invariant "SINGLE_EXECUTABLE implies RunExtract early-returns" is load-bearing. A comment tying the two would help future changes avoid breaking it.

## RSBS overlays — verified against the tree

Ran through the claims in the PR body:

- `RSBS_SINGLE_EXECUTABLE` early-return at top of ctor — **present**.
- `RunExtract`'s `sohFast3dWindow == nullptr` guard under `#ifdef RSBS_SINGLE_EXECUTABLE` — **present**.
- `SINGLE_EXECUTABLE_BUILD` freeze/resume block at file tail — unchanged (still at lines 2959–3045).
- `RsbsFeatureEnabled("RSBS_DISABLE_SHIPINIT")` wrapping `ShipInit::InitAll()` at end of `InitOTR` — **present**. ShipInit move to end of InitOTR matches upstream's move.
- `RsbsFeatureEnabled("RSBS_DISABLE_AUDIO")`, `...RANDO` — preserved.
- `OoT_gBuildVersionMajor/Minor/Patch` — **prefixed** correctly in ctor.
- `OoT_gPlayState` / `OoT_Flags_*` — untouched in this diff (rest of file still uses `OoT_gPlayState` throughout).

Count of `RSBS_DISABLE_*` guards currently in `OTRGlobals.cpp`: 10. PR body says 11 — likely an off-by-one. Not a real gap.

## Test plan — additional items I would add

On top of the PR's existing checklist:
- **Single-exe build with no `soh.o2r`** — does the ctor survive `InitResourceManager({ portArchivePath }, ...)` on a bogus path before the `RSBS_SINGLE_EXECUTABLE` early-return kicks in?
- **Corrupt `oot.o2r` reading `{0,0,0}` for version** — confirm what happens given the silent-delete concern above.
- **Kill the process during extraction** — the `threadPool` shared_ptr goes out of scope on exit; confirm no hang.
- **Each `RSBS_DISABLE_*` env-var toggle** individually, not just the three in the checklist — with 10 guards in the file, a regression in any of them wouldn't be caught by spot-checking three.

## Style nits (non-blocking)

- `totalExtract > 0.0f` on an `std::atomic<size_t>` implicitly converts to `float`. Should be `totalExtract > 0`.
- `(void)extractCount; (void)totalExtract;` in `CallZapd` leaves two unused params in the ABI. Since ZAPD uplift is deferred, a `// TODO(#<followup-issue>): wire through when ZAPD exposes zapd_report` would help future spelunkers.
- `std::dynamic_pointer_cast<Fast::Fast3dWindow>(OTRGlobals::Instance->context->GetWindow())` in `RunExtract` duplicates the file-scope `sohFast3dWindow` — could just use the global directly.

---

Net: mergeable after the `tfile` scope fix and the silent-deletion UX decision. The rest can land as follow-ups, but flagging here so they don't get lost.

spencerduncan (commented):

# Review — PR #250: port ImGui-driven extraction flow from upstream SOH (#5892)

**Verdict:** Request changes. Two real regressions in the extraction error path; the rest is nicely done.

---

## Summary

The port maps 1:1 onto upstream SOH #5892's shape: `OTRGlobals::RunExtract` drives a `while (!extractDone)` GUI loop with a state machine (`ExtractSteps`, `WindowsSteps`, `PromptSteps`), a `BS::thread_pool(1)` for the long-running ZAPD call, `std::atomic<bool> extracting` + `std::atomic<size_t> extractCount/totalExtract` for cross-thread state, and `SohGui::RegisterPopup` everywhere an SDL message box or blocking dialog used to live. Fork-specific additions (`RSBS_SINGLE_EXECUTABLE` guard, `OoT_gBuildVersion*` comparisons, `Ship::Context::LocateFileAcrossAppDirs`) are correctly applied and namespaced.

A few nice incidental fixes are bundled in: `ImGui::EndPopup()` is now correctly scoped inside `if (BeginPopupModal(...))` in `SohModals.cpp` (previous code violated the ImGui contract), the modal is centered via `SetNextWindowPos`, and the `ImFont*` members are default-initialized to `nullptr` with a guard in `Menu::DrawElement` so we can no longer crash rendering a menu before fonts load.

But the Extractor side regressed in two places that matter.

---

## Must-fix before merge

### 1. `fclose(NULL)` on ROM-permission error path — `OTRGlobals.cpp`, WS_PERMS case

```cpp
FILE* tfile = fopen("./text.txt", "w");
...
if (tfile == NULL || error) {
    SohGui::RegisterPopup("SoH Permissions Error", "...", "OK", "", [&]() {
        fclose(tfile);          // <-- UB when tfile == NULL (the reason we're in this branch)
        PathTestCleanup(tfile);
        exit(0);
    });
}
```

We only enter this branch when `tfile == NULL || error`. `fclose(NULL)` is undefined behavior per C11 7.21.5.1 — glibc happens to tolerate it, but MSVC CRT aborts. Guard it:

```cpp
if (tfile != NULL) fclose(tfile);
```

### 2. Exception handling removed from `CallZapd` — `Extractor/Extract.cpp`

The previous `CallZapd` wrapped `zapd_main(argc, argv)` in `try / catch (const std::exception&) / catch (...)` and on failure: showed an error box, hid the Windows console, restored cwd, and `remove_all(tempdir)`. The new code just calls:

```cpp
// NOTE: progress reporting via zapd_report is not yet wired up...
zapd_main(argc, (char**)argv.data());
```

Two concrete problems:

- `CallZapd` now runs inside `threadPool->detach_task([&]() { extract.CallZapd(...); ... })`. An uncaught exception on a detached task calls `std::terminate` — the process dies with no modal, no cleanup.
- Even on clean termination the old cleanup (`current_path(curdir)`, `remove_all(tempdir)`) is gone. If ZAPD exits via a non-throwing failure path the tempdir is leaked.

Restore the try/catch and surface failures through `SohGui::RegisterPopup` instead of the removed `ShowErrorBox`:

```cpp
try {
    zapd_main(argc, (char**)argv.data());
} catch (const std::exception& e) {
    SohGui::RegisterPopup("Extraction Failed",
                          std::string("ZAPD extraction failed: ") + e.what());
    std::filesystem::current_path(curdir);
    std::filesystem::remove_all(tempdir);
    return false;
} catch (...) {
    SohGui::RegisterPopup("Extraction Failed", "ZAPD extraction failed with an unknown error.");
    std::filesystem::current_path(curdir);
    std::filesystem::remove_all(tempdir);
    return false;
}
```

Bonus: since we're on the worker thread, `RegisterPopup` needs to be thread-safe — skimming `SohModalWindow::RegisterPopup` it just `push_back`s onto `modals`, which is read from the main thread in `DrawElement`. That's a race. Either lock it or post the popup from the main thread after `extracting = false`. Worth confirming even independently of this PR.

---

## Should-fix

### 3. Progress popup is visibly stuck at "Starting Up"

The comment is honest about it:

```cpp
// NOTE: progress reporting via zapd_report is not yet wired up in this tree; the ZAPD submodule
// pinned here still exposes only zapd_main. The ImGui progress popup owned by RunExtract will
// show an indeterminate "Starting Up" state until ZAPD is uplifted.
```

Two small things would soften the UX until the ZAPD uplift lands:

- Make the ImGui progress bar `ImGui::ProgressBar(-1.0f * (float)ImGui::GetTime(), ...)` for the indeterminate animation instead of a static 0.0.
- Keep a cursor on what's blocking this — drop a `// TODO(#xxx)` next to `(void)extractCount; (void)totalExtract;` in `CallZapd` so it's greppable when we pull the ZAPD update.

### 4. `[&]` capture of `tfile` outlives its scope

In WS_PERMS the popup callback captures `tfile` by reference. `FILE* tfile` is a local inside the `case { … }` block; the popup callback fires on a later frame after the case block has exited. By then `tfile` is a dangling stack reference. The callback only does `fclose(tfile); PathTestCleanup(tfile); exit(0);` and exits immediately, so this is unlikely to bite in practice, but it's UB and the `PathTestCleanup(tfile)` arg is unused anyway. Simplest fix: capture by value (`[tfile]`) and drop the NULL issue by `if (tfile) fclose(tfile);`.

---

## Nits / follow-ups

- **Fragile two-step init**: `SetupMenu()` and `SetupMenuElements()` must be called in order; forgetting the second produces an empty menu silently. A one-line doc comment on the `SetupMenu` declaration in `SohGui.hpp` explaining the split (and why — needed so `OTRGlobals::Initialize` can call `RunExtract` after the modal window exists but before menu elements are populated) would save the next person a debugging session.
- **`Menu::DrawElement` silent null-font return**: Fine as defensive code, but if fonts genuinely fail to load we'll render nothing with no signal. Consider a one-shot `SPDLOG_WARN` the first time the guard trips.
- **`extern std::shared_ptr<SohGui::SohMenu> mSohMenu;`** reaching into SohGui's internals from `OTRGlobals.cpp` — cleaner as a `SohGui::GetSohMenu()` accessor. Not a blocker.
- **`OoT_` prefix**: Correctly applied on the three `gBuildVersion*` comparisons I spot-checked. Good.
- **`BS::thread_pool(1)`**: Single-thread pool, so there's no second-extraction race — good call, matches upstream.
- **Platform guards**: SWITCH/WIIU/`_WIN32` branches look intact and the `RSBS_SINGLE_EXECUTABLE` early-return in `RunExtract` is the right fork-specific addition.

---

## Verdict

**REQUEST CHANGES.** Fixes 1 and 2 are small and should land before merge. 3 and 4 can ride along or follow up. CI is green and the upstream port is otherwise clean — once the extraction error paths are restored this is a good merge.

PR #249, opened 2026-04-21
Port Scrolling Texture Interpolation from SOH (#234)

spencerduncan:
## Summary

Ports upstream SOH commit [d4d3e8bc0](https://github.com/HarbourMasters/Shipwright/commit/d4d3e8bc0) ([HarbourMasters/Shipwright#6224](https://github.com/HarbourMasters/Shipwright/pull/6224)) — scrolling texture interpolation — into this tree. Closes deferred item from [#226](https://github.com/spencerduncan/redshipblueship/pull/226) and resolves [#234](https://github.com/spencerduncan/redshipblueship/issues/234).

Scrolling textures (Lost Woods water, Zora's Domain, Water Temple currents, fire/lava effects, boss textures, etc.) now interpolate smoothly at high framerates instead of stepping once per 20 Hz game tick.

### Core wire-up
- `games/oot/soh/OTRGlobals.cpp`: in `RunCommands`, reset `Interpreter::mInterpolationIndex` to 0 before the mtx-replacement draw loop and increment it per sub-frame.
- `games/oot/soh/OTRGlobals.{cpp,h}`: expose `Ship_GetInterpolationFPS()` and `Ship_GetInterpolationFrameCount()` (`ceil(fps/20)`) as `extern "C"` so C game code can ask how many interpolated sub-frames to emit.
- `games/oot/soh/GbiWrap.cpp`: add the `gDPSetTileSizeInterp` wrapper that emits the `G_SETTILESIZE_INTERP` opcode with float tile coordinates.
- `games/oot/include/functions.h`: declare `gDPSetTileSizeInterp`, `OoT_Gfx_TexScrollEx`, `OoT_Gfx_TwoTexScrollEx`, `OoT_Gfx_TwoTexScrollEnvColorEx`.

### New RCP helpers (`games/oot/src/code/z_rcp.c`)
`OoT_Gfx_TexScrollEx` / `OoT_Gfx_TwoTexScrollEx` / `OoT_Gfx_TwoTexScrollEnvColorEx` take per-axis step deltas and generate one `gDPSetInterpolation` + `gDPSetTileSizeInterp` pair per interpolated sub-frame, advancing the tile origin by `step/frames` each iteration so the renderer has fractional tile positions to interpolate across. The existing non-`Ex` helpers are kept for compatibility.

### Call-site updates
[...]

PR #247, opened 2026-04-21
Port Custom Messages → Hooks/ShipInit refactor from upstream SOH (#228)

spencerduncan:
## Summary
Ports upstream Shipwright commit [7b3efb1e7](https://github.com/HarbourMasters/Shipwright/commit/7b3efb1e7b85783edd5d736eff07fbeae76a0d19) ([HarbourMasters/Shipwright#5101](https://github.com/HarbourMasters/Shipwright/pull/5101)) — the *Custom Messages → Hooks + ShipInit* refactor.

- Introduces the `OnOpenText` GameInteractor hook, intercepted by ShipInit-registered handlers instead of a monolithic `Randomizer::CreateCustomMessages()` call.
- Removes ~541 lines from [OTRGlobals.cpp](games/oot/soh/OTRGlobals.cpp) and ~1800 lines of message/hint/rupee scaffolding from [randomizer.cpp](games/oot/soh/Enhancements/randomizer/randomizer.cpp), moved into 9 new files under [Enhancements/randomizer/Messages/](games/oot/soh/Enhancements/randomizer/Messages/) (ItemMessages, MerchantMessages, Navi, Rupees, Goron, GossipStoneHints, EntranceHints, StaticHints, Miscellaneous).
- Splits other inlined enhancements out of OTRGlobals: [InjectItemCounts.cpp](games/oot/soh/Enhancements/InjectItemCounts.cpp), [Items/BetterBombchuShopping.cpp](games/oot/soh/Enhancements/Items/BetterBombchuShopping.cpp), [NoSkulltulaFreeze.cpp](games/oot/soh/Enhancements/NoSkulltulaFreeze.cpp), [TimeSavers/MarketSneak.cpp](games/oot/soh/Enhancements/TimeSavers/MarketSneak.cpp), [TimeSavers/QuitFishingAtDoor.cpp](games/oot/soh/Enhancements/TimeSavers/QuitFishingAtDoor.cpp), [randomizer/BigPoes.cpp](games/oot/soh/Enhancements/randomizer/BigPoes.cpp).
- Adds per-item articles on `Item` entries and uses them in Get-Item message formatting.
- Preserves `InitOTR` ordering (`ShipInit::InitAll` → `Rando::StaticData::InitHashMaps`, from commit `9ab58a8d0`).

### RSBS adaptations
[...]

PR #266, opened 2026-04-27
Reduce Windows CI time (#177)

spencerduncan:
Closes work tracked by #177 (does not close the issue — leaving open until follow-ups land).

## Timing breakdown (current `main`)

Average of 3 most recent successful `build-windows` runs on `main`
(run IDs `24916122335`, `24912636601`, `24699295115`):

| Step | Avg duration | % of job |
| --- | --- | --- |
| Set up job | 3s | 0.2% |
| Install dependencies (choco ninja+llvm) | 12s | 0.7% |
| Git Checkout | 22s | 1.3% |
| Configure sccache | 16s | 0.9% |
| Restore Cached VCPKG | 25s | 1.5% |
| Configure Developer Command Prompt | 16s | 0.9% |
| Download soh.o2r | 1s | 0.1% |
| **Build SoH (configure + compile + link + cpack)** | **24:23** | **88.4%** |
| Unzip package | 14s | 0.8% |
| Upload build | 13s | 0.8% |
| Save Cache VCPKG | 60s | 3.6% |
| Post Configure sccache | 39s | 2.4% |
| Post Git Checkout | 4s | 0.2% |
| **Total job** | **~28:10** | |

Run-to-run jitter on `Build SoH`: **20:32 → 29:25** (~9 min spread). Likely sccache hit-rate variance, but invisible because everything was bundled into one step and no `sccache --show-stats` was emitted.

## What's already in place (so this PR doesn't repeat)

- `USE_LLD_LINKER=ON` — lld-link already used (#224 era)
- `/MP` — already set in `games/oot/CMakeLists.txt:515,531`, but no-op under Ninja (Ninja parallelizes itself)
- `--parallel 10` — already wired through; `windows-latest` is 4-core, so already over-subscribed
- vcpkg binary cache via `x-gha` — already enabled (#224)
- sccache cap 5G — already in place (#224)
- Release builds skip `/Zi` so sccache works (`CMake/DefaultCXX.cmake:13-19`)
- `docs-bypass.yml` already short-circuits docs-only PRs

### 1. Split the monolithic Build SoH step + emit sccache stats (diagnostic)

The 24-minute "Build SoH" step is now four steps: `CMake configure (Windows)` / `CMake build (Windows)` / `sccache stats (Windows)` / `Package (Windows)`. Per-stage timing shows up in the Actions UI for free, and the cache hit ratio becomes measurable instead of inferred. Directly addresses the "Data to collect" checklist in #177.

Direct savings: 0. **Required prerequisite for any further optimization, including measuring the impact of this PR.**

### 2. Save sccache on PR branches, not just `main` (real perf)

Removed `save: ${{ github.ref_name == github.event.repository.default_branch }}` from the sccache configure step. Previously PR branches restored from main but never persisted their own state, so every push to the same PR cold-started from main's cache. Phase 2's Windows-only `int-switch-oot-mm` / `int-switch-mm-oot` integration tests will push to the same PR many times; with save enabled, push N+1 picks up sccache from push N.

Expected: **subsequent PR pushes drop from ~24 min Build SoH to ~5 min** (link- and cpack-dominated). First push to a fresh branch and `main` builds are unchanged.

Also fixed a stray `set ` prefix on `$env:PATH=...` that was a PowerShell no-op (carried over unchanged into the new Configure step).

## Diff

```yaml
# .github/workflows/generate-builds.yml — build-windows job

  - Configure sccache: dropped `save: <main-only>` guard (defaults to save: true)
  - Build SoH (single step, ~24 min) split into:
      - CMake configure (Windows)
      - CMake build  (Windows)
      - sccache stats (Windows)         # if: always()
      - Package (Windows)                # cpack + zip rename
```

1 file changed, 22 insertions(+), 6 deletions(-).

## Before / after table — to be filled in once CI runs land

| Scenario | Before (main) | After (this PR, run 1) | After (this PR, run 2 — same branch) |
| --- | ---: | ---: | ---: |
| Total `build-windows` job | ~28:10 | _tbd_ | _tbd_ |
| `CMake configure` | (bundled) | _tbd_ | _tbd_ |
| `CMake build` | (bundled) | _tbd_ | _tbd_ |
| `Package` (cpack) | (bundled) | _tbd_ | _tbd_ |
| sccache hit rate | unknown | _tbd_ | _tbd_ |

**Validation plan:**
1. First run after merge: same speed as today (cold against main's sccache); confirms no regression and gives baseline per-stage numbers.
2. Push an empty/whitespace commit to the same branch: should be drastically faster (sccache hit on most TUs); validates change #2.
3. Update the table in this PR description once both runs are complete.

## Follow-ups (separate PRs, do not block this one)

- **Remove `/FORCE:MULTIPLE`** in `games/oot/CMakeLists.txt:547,562`. Currently masks duplicate-symbol bugs that surface on Linux (see history of #247, #249). Switching to lld-link without it requires fixing those root causes first; this PR keeps the flag deliberately. Tracked separately.
- **Skip `Save Cache VCPKG`** when `steps.restore-cache-vcpkg.outputs.cache-hit == 'true'`. Saves ~60s per `main` run. Trivial diff, but skipped here to keep this PR tightly scoped to the two highest-leverage items.
- Once data from change #1 lands, revisit whether vcpkg install or compile dominates and target the larger of the two.

## Test plan

- [ ] PR triggers `generate-builds` — confirm `build-windows` succeeds with new step layout.
- [ ] Inspect the new per-step timings in the Actions UI; record numbers in the table above.
- [ ] Inspect `sccache stats (Windows)` step output; record cache hit ratio.
- [ ] Push a no-op follow-up commit; confirm second run on the same branch is materially faster (validates change #2).
- [ ] Confirm `rsbs-windows.zip` artifact is still produced and matches main's structure.

Do NOT merge until the validation pushes complete and the table is filled in.

https://claude.ai/code/session_01BhYPie7ZL6DxsEWPUnGvCU

---
_Generated by [Claude Code](https://claude.ai/code/session_01BhYPie7ZL6DxsEWPUnGvCU)_

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6668966481.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6668817395.zip)

### DIFF
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -354,7 +354,10 @@ jobs:
       with:
         variant: sccache
         max-size: "5G"
-        save: ${{ github.ref_name == github.event.repository.default_branch }}
+        # Save on every branch (default), not just main, so subsequent pushes to
+        # the same PR pick up the previous push's cache instead of cold-restoring
+        # from main. Issue #177 — Phase 2 iteration on Windows-only integration
+        # tests will rebuild many times per PR.
         key: ${{ runner.os }}-sccache-${{ github.ref }}
         restore-keys: |
           ${{ runner.os }}-sccache-${{ github.ref }}
@@ -376,15 +379,28 @@ jobs:
       with:
         name: soh.o2r
         path: bw/soh
-    - name: Build SoH
+    # Build is split into Configure / Build / Package so the GitHub Actions UI
+    # surfaces per-stage timing. Issue #177 — previously a single ~24 min step,
+    # making it impossible to tell whether time was going to vcpkg, compile, link
+    # or cpack. Keep these as separate steps even after tuning is done; the cost
+    # is negligible (~1s of step overhead) and the diagnostic value is permanent.
+    - name: CMake configure (Windows)
       env:
         VCPKG_ROOT: ${{github.workspace}}/vcpkg
       run: |
-        set $env:PATH="$env:USERPROFILE/.cargo/bin;$env:PATH"
+        $env:PATH="$env:USERPROFILE/.cargo/bin;$env:PATH"
         cmake -S . -B bw -G Ninja -DCMAKE_MAKE_PROGRAM=ninja -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DBUILD_REMOTE_CONTROL=1 -DREDSHIP_BUILD_SHARED=ON -DUSE_LLD_LINKER=ON
-        cmake --build bw --config Release --parallel 10
-
-        (cd bw && cpack)
+    - name: CMake build (Windows)
+      env:
+        VCPKG_ROOT: ${{github.workspace}}/vcpkg
+      run: cmake --build bw --config Release --parallel 10
+    - name: sccache stats (Windows)
+      if: always()
+      run: sccache --show-stats
+    - name: Package (Windows)
+      run: |
+        cd bw
+        cpack
         cd ..
         mv _packages/*.zip _packages/rsbs-windows.zip
     - name: Unzip package


spencerduncan (commented):
# PR #266 review — Reduce Windows CI time (#177)

**Verdict: Approve.** Tightly-scoped, conservative, well-instrumented. Plan-first instruction was followed: the body opens with a 3-run timing breakdown (run IDs cited), explicitly enumerates what's already in place to avoid duplication, picks two highest-leverage changes, and defers everything else to follow-ups. `build-windows` check is green (`SUCCESS`).

**Cited expected savings:** *"subsequent PR pushes drop from ~24 min Build SoH to ~5 min"* — i.e. **~19 min per iterative push** after the first build on a branch. First push on a fresh branch and `main` builds are unchanged.

---

## Priority checks

### Did it plan first? — Yes
PR body contains exactly the structure requested:
- **(a)** Timing breakdown of existing Windows run — full table averaging 3 recent `main` runs (`24916122335`, `24912636601`, `24699295115`), with `Build SoH` clocked at 24:23 (88.4% of the 28:10 total) and a documented run-to-run jitter of 20:32 → 29:25.
- **(b)** Two changes applied: (1) split monolithic `Build SoH` into Configure / Build / sccache stats / Package; (2) drop the main-only guard on sccache `save:`.
- **(c)** Before/after expectation: ~24 min → ~5 min on subsequent same-branch pushes (link- and cpack-dominated). Before/after table is included with placeholders to be filled in post-merge — explicit gate on merge until validation pushes complete.

### `/FORCE:MULTIPLE` — Untouched ✓
Only one file changed: `.github/workflows/generate-builds.yml` (22+/6-). `games/oot/CMakeLists.txt` is **not** in the changeset. Body explicitly calls out that `/FORCE:MULTIPLE` removal is a deferred follow-up because it would surface duplicate-symbol bugs (history of #247, #249) that need to land first. Instruction respected.

### Caching changes — Sound
The only caching change is dropping `save: ${{ github.ref_name == github.event.repository.default_branch }}` on the sccache action. Hashes/keys themselves are unchanged:
- `key: ${{ runner.os }}-sccache-${{ github.ref }}` — for `pull_request` events `github.ref` is `refs/pull/N/merge`, stable across pushes to the same PR. ✓
- Restore-keys fall back to `refs/heads/main` then bare `Windows-sccache`. ✓
- Net effect: push N+1 to a PR restores from push N, instead of cold-restoring from main every time. This is exactly the behavior the body claims.
- vcpkg cache (`Save Cache VCPKG`) is **deliberately not** flipped to all-branches — body explains it's a separate ~60s/run follow-up, kept out to keep this PR scoped. Good discipline.

**Minor caveat (non-blocking):** GHA per-repo cache budget is 10 GB with LRU eviction. With a 5 G sccache cap, a small number of in-flight PRs + main + vcpkg can drive eviction. Worth watching once Phase 2 ramps up; not a reason to block.

### Job split — Artifacts threaded correctly
The `Build SoH` step splits into four steps but the produced filesystem artifact is identical: `bw/` is left in place, `_packages/*.zip` is renamed to `_packages/rsbs-windows.zip` in the new `Package` step, then `Unzip package` and `Upload build` consume that path unchanged. No artifact name or path renames. ✓ The `sccache stats (Windows)` step has `if: always()` so cache-stat output is captured even on build failure — correct intent.

### Skipping / false-green risk — None introduced
- No `continue-on-error` added.
- No `paths-filter` or new conditional `if:` that would skip the build.
- The new `if: always()` is on the *stats* step only, which has no side-effect on the artifact.
- The `build-windows` check on `headRefOid 6754b324` is `SUCCESS`, confirming the new step layout produces the expected zip.

---

## Other observations

### Bug fix bundled in (good catch)
`set $env:PATH="..."` → `$env:PATH="..."`. The leading `set` was being parsed as the `Set-Variable` alias against a literal `=...` argument and was a no-op; the cargo bin path was never actually being prepended on Windows. Worth a one-line callout in the body, but the change is correct.

### Very minor — error chaining in `Package`
The original `(cd bw && cpack)` short-circuited via pwsh 7's `&&`. The new step uses `cd bw; cpack; cd ..; mv _packages/*.zip ...` (sequential). If `cpack` fails non-fatally, `cd ..` and `mv` run anyway and the step still surfaces a failure (because `mv` errors with no zip), so behavior is effectively equivalent. Not worth changing.

### Validation gate is appropriate
The body explicitly states *"Do NOT merge until the validation pushes complete and the table is filled in."* That's the right call — change #2's expected savings are unobservable without the per-step timings produced by change #1 (and a same-branch repush). Approving the code, not the merge.

---

## Suggestions (non-blocking)

1. After merge: confirm sccache cache scope behavior matches expectation by checking that two consecutive pushes to the same PR show a `cache hits` jump in the new `sccache stats (Windows)` step output.
2. Once data lands, the follow-up to skip `Save Cache VCPKG` on cache-hit (60s/main run) is genuinely trivial and worth picking up immediately.
3. The cited 24 → 5 min figure is the optimistic case (cache-hot link-dominated rebuild). Recommend the table also capture a "pure no-op recompile" run separately from a "1-file-edit recompile" run, so the steady-state Phase 2 iteration cost is well-characterized before integration tests start churning the same Windows job.

PR #267, opened 2026-04-27
Mirror OoT extern SaveContext on MM side (follows up #240)

spencerduncan:
## Summary

PR #240 added an explicit `extern SaveContext gSaveContext;` declaration to the OoT TU at `games/oot/soh/GameExports_SingleExe.cpp` so that `OoT_Game_Resume` could reliably restore the SaveContext on return from MM (issue #170). The MM TU at `games/mm/2s2h/GameExports_SingleExe.cpp` was not updated in the same way, even though `MM_Game_Resume` performs the equivalent restore (also added in #240).

Today the MM TU builds because the symbol is pulled in transitively via `extern "C" { ... #include "global.h" }` → `games/mm/include/z64save.h:1881`, which contains `extern SaveContext gSaveContext;`. This works by accident.

## Why it matters

If `global.h` ever stops including `z64save.h`, or the include order shifts (e.g. someone reorganises the header pyramid as part of an MM cleanup), `MM_Game_Resume`'s restore path silently loses the declaration and fails to compile — and the breakage points at MM_Game_Resume rather than at the real cause. Mirroring OoT's explicit extern makes MM's view of `gSaveContext` independent of header transitivity, matching the pattern already established for OoT.

## Change

One six-line block added inside the existing `extern "C" {}` declarations block in `games/mm/2s2h/GameExports_SingleExe.cpp`, immediately after the audio-related externs and before the screen-dimension globals — the same relative position as the OoT mirror.

## Verification

Relying on CI to validate the build (no local cmake available on author's machine). No behaviour change is expected; this is a pure declaration hardening that mirrors an already-merged OoT change.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6671686605.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6671503716.zip)

### DIFF
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -66,6 +66,12 @@ extern "C" {
     extern s32 gAudioCtxInitalized;
     void AudioThread_InitMesgQueues(void);
 
+    // MM's SaveContext (type defined via global.h -> z64save.h).
+    // Declared here so MM_Game_Resume() can restore it on return from OoT (#170).
+    // Mirrors the explicit extern on the OoT TU — keeps the symbol's visibility
+    // independent of whether global.h ever stops including z64save.h transitively.
+    extern SaveContext gSaveContext;
+
     // Globals from main.c
     extern s32 MM_gScreenWidth;
     extern s32 MM_gScreenHeight;


spencerduncan (commented):
## Verdict: APPROVE-with-comments

The one-line addition is harmless and CI is green across all jobs (clang-tidy, clang-format, build-linux, build-windows, integration-tests-linux, generate-rsbs-otr). Re-declaring `gSaveContext` with the same type alongside the transitive declaration from `z64save.h` is well-defined and adds no runtime behavior. No blockers.

That said, the PR description and the change's stated rationale don't quite line up with what's in the tree. None of the comments below are merge-blocking — the code itself is fine.

---

### 1. The "MM_Game_Resume performs the equivalent restore" claim doesn't match the current code

The PR body says:

> The MM TU at `games/mm/2s2h/GameExports_SingleExe.cpp` was not updated in the same way, even though `MM_Game_Resume` performs the equivalent restore (also added in #240).

But on this branch `MM_Game_Resume` is still a TODO stub:

```cpp
void MM_Game_Resume(void) {
    fprintf(stderr, "[MM] Game_Resume called\n");
    fflush(stderr);
    // TODO: Restore MM audio, reload MM-specific resources if needed
}
```

`gSaveContext` does not appear anywhere else in this TU (a grep across `games/mm/2s2h/GameExports_SingleExe.cpp` finds no other references). So this PR adds a forward declaration for a symbol that has zero use sites in the file. The hardening it's framed as is forward-looking rather than current — there is presently no `MM_Game_Resume` body for `extern SaveContext gSaveContext;` to enable.

That makes the change harmless but also currently inert. Two ways to make it land cleanly:

- **(a)** Reframe the PR description as "stage the declaration ahead of the upcoming MM-side restore work" rather than describing existing behavior — clear about intent, easier to review later. Or
- **(b)** Bundle this with the actual `MM_Game_Resume` body that uses `gSaveContext`, so the declaration has a real consumer the moment it lands. This also lets reviewers verify the declaration is in scope at the use site, which is the failure mode #240 originally hit.

Either is fine; (b) is closer to the spirit of "follows up #240" since #240's value came from the use site, not the declaration in isolation.

### 2. The "same relative position as the OoT mirror" claim is also off

The PR body says the new extern sits "in the same relative position as the OoT mirror." On the OoT side, `gSaveContext` is declared at [GameExports_SingleExe.cpp:237](games/oot/soh/GameExports_SingleExe.cpp:237) — in the cross-game-entrance block at the bottom of the file, immediately above its only use site at [GameExports_SingleExe.cpp:286](games/oot/soh/GameExports_SingleExe.cpp:286) inside `Combo_CheckEntranceSwitch`:

```cpp
// OoT's SaveContext (type defined via OTRGlobals.h -> z64save.h)
extern "C" SaveContext gSaveContext;
...
Combo_FreezeState("oot", returnEntrance, &gSaveContext, sizeof(gSaveContext));
```

MM places the new declaration in the upper main `extern "C" { … }` block at [GameExports_SingleExe.cpp:71](games/mm/2s2h/GameExports_SingleExe.cpp:71), between the audio externs and the screen-dimension globals — structurally and positionally different. Functionally fine (declaration order is satisfied since there's currently no use site at all), but the description shouldn't claim positional parity. Worth tightening the wording — small thing, but it makes future reviewers waste time looking for a parallel that isn't there.

### 3. Style nit (non-blocking): inline vs. block form

OoT uses the inline form:

```cpp
extern "C" SaveContext gSaveContext;
```

MM uses the block form (declaration inside an existing `extern "C" { … }`):

```cpp
extern "C" {
    ...
    extern SaveContext gSaveContext;
    ...
}
```

Both are valid and link-compatible. The MM block form arguably matches the surrounding MM style (most externs in this file already live inside the existing block), so this is fine as-is — flagging only for consistency awareness if you ever decide to standardize.

### 4. Transitive-include landmine coverage

The hardening rationale (insulating against `global.h` ever dropping its `z64save.h` include) is sound *in principle*, but until there's a `gSaveContext` use site in this TU, removing that transitive include wouldn't break anything here either way — there's nothing to fail to compile. Once `MM_Game_Resume` (or some other function in this file) actually references `gSaveContext`, the hardening becomes load-bearing. Worth keeping that pairing in mind for the follow-up work.

There may be other transitive-include landmines in the same neighborhood (e.g., types pulled in indirectly via the `extern "C" { #include "global.h" }` umbrella), but those are out of scope for a PR this targeted — and the right time to surface them is when each one gets a use site added.

---

**Bottom line:** code change is fine, ship it. Just sharpen the PR description (or land it with the actual `MM_Game_Resume` body) so the rationale matches the code.

PR #268, opened 2026-04-27
Thread-safety guard on SohModalWindow::RegisterPopup (follows up #250)

spencerduncan:
## Summary

PR #250 ("ImGui-Driven Extraction Flow") introduced four `threadPool->detach_task` bodies inside `OTRGlobals::RunExtract` (around lines 471/480/534/554 of the merged diff) that call `Extractor::CallZapd` from a worker thread. PR #250's commit message explicitly investigated whether `SohModalWindow::RegisterPopup` races with those workers and concluded that **today** it does not — the four detach_task bodies surface errors via native `MessageBoxA` / `SDL_ShowSimpleMessageBox` in `Extractor::ShowErrorBox`, not `SohGui::RegisterPopup`. All RegisterPopup callsites today run on the main state-machine thread.

So the queue is currently race-free, but only by accident. Any future detached worker that decides to call `SohGui::RegisterPopup` (or `mModalWindow->RegisterPopup` directly) would silently introduce a data race against `DrawElement` on the main thread, with no compile-time or runtime warning.

## Change

Adds a `static std::recursive_mutex modalsMutex` to `games/oot/soh/SohGui/SohModals.cpp` covering the `modals` queue and the `closePopup` flag. Every public entry point on `SohModalWindow` now takes the lock: `DrawElement`, `RegisterPopup`, `IsPopupOpen`, `PopupsQueued`, `DismissPopup`.

The mutex is recursive because `DrawElement` invokes user-supplied button callbacks while holding the lock, and several existing callsites legitimately register another popup from inside one of those callbacks (the most prominent example is the "Confirm Re-extract" path in `OTRGlobals::RunExtract` itself, which inside its button1callback calls `threadPool->detach_task` whose body then doesn't re-enter — but the recursive pattern also covers e.g. `mod_menu.cpp`'s "Apply & Close" flow). A non-recursive mutex would deadlock those flows.

## Why mutex (not "main thread only" assert)

The user's audit asked for the simpler/safer of: (a) mutex, or (b) `assert(IsMainThread())` plus a documentation block. The codebase has no main-thread tracking primitive to assert against (`grep -rn IsMainThread games/oot/soh` returns nothing), so option (b) would degrade to a comment-only fix that the type system can't enforce. The mutex covers a small, low-frequency surface (one push_back per popup, one frame's worth of reads per draw) and actually enforces the invariant.

## Scope

Touches only `games/oot/soh/SohGui/SohModals.cpp`. The MM-side equivalent (`games/mm/2s2h/BenGui/BenModals.cpp`) has the same shape but is not currently exposed to detach_task callers — leaving that for a separate follow-up if/when MM gains worker-thread popup paths.

## Verification

Relying on CI to validate the build (no local cmake available on author's machine).

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6672189252.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6672057625.zip)

### DIFF
--- a/games/oot/soh/SohGui/SohModals.cpp
+++ b/games/oot/soh/SohGui/SohModals.cpp
@@ -1,5 +1,6 @@
 #include "SohModals.h"
 #include <imgui.h>
+#include <mutex>
 #include <vector>
 #include <string>
 #include <libultraship/bridge.h>
@@ -22,6 +23,14 @@ std::vector<SohModal> modals;
 
 bool closePopup = false;
 
+// Guards the `modals` queue and `closePopup` flag. Recursive because
+// DrawElement invokes user-supplied button callbacks while holding the
+// lock, and several callsites legitimately register another popup from
+// inside one of those callbacks (e.g. OTRGlobals.cpp's "Confirm
+// Re-extract" path). Added so any future thread-pool worker that calls
+// RegisterPopup can do so safely against the main-thread DrawElement.
+static std::recursive_mutex modalsMutex;
+
 void SohModalWindow::Draw() {
     if (!IsVisible()) {
         return;
@@ -32,6 +41,7 @@ void SohModalWindow::Draw() {
 }
 
 void SohModalWindow::DrawElement() {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     if (modals.size() > 0) {
         SohModal curModal = modals.at(0);
         if (!ImGui::IsPopupOpen(curModal.title_.c_str())) {
@@ -76,17 +86,21 @@ void SohModalWindow::DrawElement() {
 
 void SohModalWindow::RegisterPopup(std::string title, std::string message, std::string button1, std::string button2,
                                    std::function<void()> button1callback, std::function<void()> button2callback) {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     modals.push_back({ title, message, button1, button2, button1callback, button2callback });
 }
 
 size_t SohModalWindow::PopupsQueued() {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     return modals.size();
 }
 
 bool SohModalWindow::IsPopupOpen(std::string title) {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     return !modals.empty() && modals.at(0).title_ == title;
 }
 
 void SohModalWindow::DismissPopup() {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     closePopup = true;
 }


spencerduncan (commented):
## Verdict: Approve, with one optional suggestion

The lock covers the right surface, all five mutating/reading entry points are guarded, the `static` file-local mutex is the correct linkage, and the cross-reference holds: none of the four `threadPool->detach_task` bodies in `OTRGlobals::RunExtract` (lines 630, 639, 693, 713 in the head version) call `RegisterPopup` today, so this is the prophylactic guard you describe. CI is green across all platforms. Safe to merge as-is.

The one piece worth pushing on is the `recursive_mutex` justification — it's a slightly stronger choice than the actual call graph requires. Details below.

---

### What I verified

**Coverage of `modals` / `closePopup`.** Five public methods touch the protected state; all five take the lock:

- `DrawElement()` — reads/writes both
- `RegisterPopup()` — writes `modals`
- `PopupsQueued()` — reads `modals`
- `IsPopupOpen()` — reads `modals`
- `DismissPopup()` — writes `closePopup`

`Draw()` doesn't touch state directly and just delegates to `DrawElement`, so leaving it unlocked is correct. `closePopup` is also read and cleared inside `DrawElement` while the lock is held — the latent race between `DismissPopup` (write) and `DrawElement` (read+clear) is now closed.

**No reference leaks out of the lock.** `DrawElement` does `SohModal curModal = modals.at(0);` — that's a copy, so callback invocation can't observe a dangling reference if `modals` is mutated. Good.

**`detach_task` body audit (PR body claim).** Confirmed. The four worker bodies at `games/oot/soh/OTRGlobals.cpp:630`, `:639`, `:693`, `:713` only call `extract.CallZapd(...)` and mutate `extracting` / `extractCount` / `totalExtract` / `extractStep` / `promptStep` / `generatedIsMQ`. None call `SohGui::RegisterPopup` or any other `SohModalWindow` method. So the queue is race-free *today* and this PR turns "race-free by accident" into "race-free by construction" — the framing in the PR description is accurate.

(Side note, out of scope: those flags are plain `bool` / `int` globals shared across the worker and the main loop. That's a separate latent race — `extracting` in particular is read at `OTRGlobals.cpp:491` from the main thread and written at `:633`/`:642`/`:697`/`:716` from workers. Worth a follow-up `std::atomic<bool>` pass, but not this PR.)

**`static` linkage.** Matches the right convention. `modals` and `closePopup` themselves have external linkage in the existing code (no `static`, no header-export — also a minor latent issue, also out of scope). Making the new mutex `static` is the *better* choice; don't change it.

---

### One suggestion (non-blocking): the recursive-mutex rationale doesn't hold up under inspection

The PR comment / description argues that `DrawElement` "invokes user-supplied button callbacks while holding the lock, and several callsites legitimately register another popup from inside one of those callbacks (e.g. `OTRGlobals.cpp`'s 'Confirm Re-extract' path … `mod_menu.cpp`'s 'Apply & Close' flow)."

I went looking for an actual button-callback that re-enters `RegisterPopup` / `DismissPopup` / `IsPopupOpen` and couldn't find one:

- **"Confirm Re-extract"** (`OTRGlobals.cpp:628`) — the body of `button1callback` is `extracting = true; threadPool->detach_task([&]() { extract.CallZapd(...); ... });`. The detach_task body runs asynchronously on a worker (which would acquire the lock fresh, no recursion), and the synchronous body itself doesn't touch `SohModalWindow`. The PR description even concedes this ("then doesn't re-enter").
- **"Apply & Close"** (`mod_menu.cpp:315`) — this `RegisterPopup` is called from a `UIWidgets::Button("Apply & Close", ...)` *inside the mod menu UI*, not from inside a `SohModalWindow::DrawElement` button callback. So it isn't re-entry through the lock; it's a fresh acquisition from a different ImGui draw context.
- All other `RegisterPopup` callsites I checked (`SaveManager.cpp:491,521,1321`, `randomizer.cpp:543`, `Menu.cpp:759`, `SohMenuSettings.cpp:421`, `OTRGlobals.cpp:1680`, `FileSelectEnhancements.cpp:91`) are also outside any modal button-callback context. The `OTRGlobals.cpp:1677-1683` F1-key handler does `IsPopupOpen` + `DismissPopup`/`RegisterPopup` sequentially, but those are separate lock acquisitions on the same thread, not nested.

So the *current* call graph would be perfectly happy with a plain `std::mutex`. The recursive choice is defensive against a future contributor adding a button callback that re-enters — which is plausible but speculative.

If you want to keep `recursive_mutex` as a footgun-prevention measure, fine — please update the comment to say that explicitly ("future-proofing against re-entrant callbacks; no current callsite re-enters") rather than pointing at "Confirm Re-extract" / "Apply & Close" as concrete examples, since neither actually exhibits the pattern.

The cleaner alternative is to release the lock around the callback invocation:

```cpp
if (ImGui::Button(curModal.button1_.c_str())) {
    auto cb = std::move(curModal.button1callback_);
    modals.erase(modals.begin());
    lock.unlock();
    if (cb) cb();
    ImGui::CloseCurrentPopup();
    return;  // re-enter next frame; avoid touching modals after unlock
}
```

That gives you a plain `std::mutex` *and* eliminates the general "holding a lock across user-supplied code" anti-pattern (priority inversion, surprise reentrancy, impossibility of doing async work under the lock). But it materially restructures `DrawElement` and isn't strictly required for correctness — happy to defer to a follow-up if you prefer to keep this PR minimal.

---

### Nits (also out of scope, listed for visibility)

- `ImGui::EndPopup()` at the end of `DrawElement` is called unconditionally even when `BeginPopupModal` returns false. Pre-existing — not introduced or worsened by this PR — but ImGui usually asserts on this. Worth a separate cleanup.
- `bool closePopup = false;` and `std::vector<SohModal> modals;` have external linkage. Should probably be `static` (or in an anonymous namespace). Pre-existing.

Neither blocks this PR.

---

**TL;DR:** Lock placement is correct, coverage is complete, the cross-thread invariant the PR claims to enforce is real and now enforced. Only ask is to fix the comment so the `recursive_mutex` rationale matches what the code actually does (defensive future-proofing, not "we have re-entrant callbacks today").

PR #269, opened 2026-04-27
Mirror SohModalWindow thread-safety on MM-side BenModalWindow (follows up #268)

spencerduncan:
## Summary

MM-side mirror of the thread-safety guard applied in #268 to `SohModalWindow::RegisterPopup`. `BenModals.cpp` has the same shape and the same lack of synchronization on the shared `modals` queue and `closePopup` flag — closing the gap so MM doesn't ship the same footgun.

The PR #268 worker explicitly flagged the MM-side as out of scope for that change; this is the follow-up.

## Change

Adds a `static std::recursive_mutex modalsMutex` and a `std::lock_guard` at the top of every public method that touches the `modals` collection or `closePopup` bool:

- `BenModalWindow::DrawElement`
- `BenModalWindow::RegisterPopup`
- `BenModalWindow::IsPopupOpen`
- `BenModalWindow::DismissPopup`

Recursive for the same reason as #268: `DrawElement` invokes user-supplied button callbacks while holding the lock, and a callback may re-enter `RegisterPopup`.

13 insertions, 0 deletions. No behavior change on the single-threaded path.

## Test plan

- [ ] CI build (Linux / Windows / macOS) succeeds
- [ ] No new warnings on the touched file

Verification: relying on CI to validate (no local cmake on author's machine).

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6672841709.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/6672716618.zip)

### DIFF
--- a/games/mm/2s2h/BenGui/BenModals.cpp
+++ b/games/mm/2s2h/BenGui/BenModals.cpp
@@ -1,5 +1,6 @@
 #include "BenModals.h"
 #include <imgui.h>
+#include <mutex>
 #include <vector>
 #include <string>
 #include "UIWidgets.hpp"
@@ -17,6 +18,14 @@ std::vector<BenModal> modals;
 
 bool closePopup = false;
 
+// Guards the `modals` queue and `closePopup` flag. Recursive because
+// DrawElement invokes user-supplied button callbacks while holding the
+// lock, and a callback may legitimately register another popup from
+// inside one of those callbacks. Added so any future thread-pool worker
+// that calls RegisterPopup can do so safely against the main-thread
+// DrawElement. Mirrors the SohModalWindow guard added in #268.
+static std::recursive_mutex modalsMutex;
+
 void BenModalWindow::Draw() {
     if (!IsVisible()) {
         return;
@@ -27,6 +36,7 @@ void BenModalWindow::Draw() {
 }
 
 void BenModalWindow::DrawElement() {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     if (modals.size() > 0) {
         BenModal curModal = modals.at(0);
         if (!ImGui::IsPopupOpen(curModal.title_.c_str())) {
@@ -70,13 +80,16 @@ void BenModalWindow::DrawElement() {
 
 void BenModalWindow::RegisterPopup(std::string title, std::string message, std::string button1, std::string button2,
                                    std::function<void()> button1callback, std::function<void()> button2callback) {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     modals.push_back({ title, message, button1, button2, button1callback, button2callback });
 }
 
 bool BenModalWindow::IsPopupOpen(std::string title) {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     return !modals.empty() && modals.at(0).title_ == title;
 }
 
 void BenModalWindow::DismissPopup() {
+    std::lock_guard<std::recursive_mutex> lock(modalsMutex);
     closePopup = true;
 }


spencerduncan (commented):
## Code review — PR #269: Mirror SohModalWindow thread-safety on MM-side BenModalWindow

**Verdict: approve.** Clean, minimal, faithful mirror of #268. No blocking issues.

Head: `7c8d1ae3`. Mergeable. All CI green (clang-tidy, clang-format, build-linux, build-windows, generate-rsbs-otr, integration-tests-linux); macOS skipped per workflow config.

### Pattern parity vs. #268

Reviewed `games/oot/soh/SohGui/SohModals.cpp` diff alongside `games/mm/2s2h/BenGui/BenModals.cpp` diff. Mirror is faithful:

| Aspect | OoT (#268) | MM (#269) | Match? |
| --- | --- | --- | --- |
| Lock type | `std::recursive_mutex` | `std::recursive_mutex` | ✅ |
| Storage | `static` file-scope, named `modalsMutex` | `static` file-scope, named `modalsMutex` | ✅ |
| Header include | `<mutex>` after `<imgui.h>` | `<mutex>` after `<imgui.h>` | ✅ |
| Lock placement | First statement in each public method | First statement in each public method | ✅ |
| Guard type | `std::lock_guard<std::recursive_mutex>` | `std::lock_guard<std::recursive_mutex>` | ✅ |
| Comment | Block above the mutex explaining recursion + future-worker rationale | Same shape, MM-flavoured | ✅ |

### Public-method coverage

Confirmed the 4-vs-5 lock-site delta is correct:

- `BenModals.h` declares **only**: `Draw`, `InitElement`, `DrawElement`, `UpdateElement`, `RegisterPopup`, `IsPopupOpen`, `DismissPopup`.
- `grep -rn "PopupsQueued" games/mm` and `games/oot` → **zero matches**. The method does not exist on either side in the current tree (PR #268 must add it on the OoT side as part of that change; it was never ported to MM and has no callers in MM). The MM mirror has nothing to lock here.
- `Draw()` is correctly **not** locked — it only calls `DrawElement()` (which holds the lock) and `SyncVisibilityConsoleVariable()`. No direct touch of shared state.
- `InitElement()`/`UpdateElement()` are header-inline no-ops (`{};`). Nothing to guard.

All 4 methods that touch `modals` or `closePopup` (`DrawElement`, `RegisterPopup`, `IsPopupOpen`, `DismissPopup`) are now locked. Coverage is complete.

### `std::recursive_mutex` choice

Same logic as #268. `BenModalWindow::DrawElement` (BenModals.cpp:46–63) invokes user-supplied `button1callback_` / `button2callback_` while the lock is held. If a callback re-enters `RegisterPopup`, a non-recursive mutex would self-deadlock. Recursive is the correct call.

One observation worth flagging (non-blocking): on OoT there is a *real* re-entrant callsite — `games/oot/soh/OTRGlobals.cpp:1468` registers the "Menu Moved" popup from inside another callback path (the `Confirm Re-extract` flow the PR body cites). On MM, the only `RegisterPopup` callsite I can find is `games/mm/2s2h/BenGui/Menu.cpp:702` (the Quit 2S2H button), and neither of its callbacks re-enters `RegisterPopup`. So today MM's recursion is purely defensive — the lock would behave identically as a plain `std::mutex`. The author's PR body already frames this as forward-looking ("any future thread-pool worker") and matching the OoT side for symmetry; that is a reasonable rationale and I would not push back on `recursive_mutex` here. Keeping the two files isomorphic has more value than shaving a few cycles.

### Worker-thread / `detach_tasks` context

Searched for `detach_tasks` across `games/mm` and `games/oot`: zero matches. There is no thread-pool plumbing today on either side that calls `RegisterPopup` off the main thread. So both #268 and #269 are purely defensive against future cross-thread callers. That matches what the PR body says ("Added so any future thread-pool worker… can do so safely") — no misrepresentation. ✅

The `std::thread` hits in MM (`BenPort.cpp` audio thread, `Extract.cpp` MessageboxWorker, `AudioSampleFactory.cpp` decoder workers) do not touch `RegisterPopup`, so they are not relevant here.

### MM-specific callbacks

Walked the only existing caller (`Menu.cpp:702`, Quit 2S2H button):
- `button1callback`: toggles menu visibility, calls `Window::Close()`. No `RegisterPopup`.
- `button2callback`: `nullptr`.

No MM-side BenGui menu items recursively invoke `RegisterPopup` today. The recursive-mutex comment block in `BenModals.cpp` is correctly worded as conditional ("a callback **may** legitimately register another popup"), unlike the OoT version which cites a concrete callsite. That phrasing is appropriate for the MM reality.

### Other notes

- 13 insertions / 0 deletions per the PR body — confirmed by the diff. No behaviour change on the single-threaded path.
- No formatting drift; clang-format CI passed.
- `static` linkage on `modalsMutex` keeps it TU-local, matching the existing `static`-feel of the file-scope `modals` vector and `closePopup` flag (those two are technically extern-default but only referenced from this TU).

### Suggestions (optional, non-blocking)

1. Long-term, the file-scope `std::vector<BenModal> modals;` and `bool closePopup = false;` are not `static` — they have external linkage. They're fine because no other TU references them, but if you ever want to harden this further, marking them `static` (or moving them into an anonymous namespace alongside `modalsMutex`) would make the locking contract structurally enforceable. Out of scope for this PR.
2. If a `PopupsQueued()` is ever added to MM in the future, remember to lock it too. (Not actionable now.)

### Final call

Approve and merge. This is the right shape for the follow-up — minimal, mirror-symmetric with #268, CI green, and the rationale in the PR body matches reality (defensive, forward-looking, no current cross-thread caller).

PR #348, opened 2026-07-15
Fix OoT Market cutscene crash from cross-game entrance leaking into entranceIndex

spencerduncan:
## Summary

Fixes the reported Windows crash: `0xC0000005` in `Cutscene_HandleConditionalTriggers` with `entrance: 49168`, scene `SCENE_MARKET_DAY` (child, day).

It's an **out-of-bounds read of OoT's `gEntranceTable`**:
- `49168 = 0xC010 = MM_ENTR_CLOCK_TOWER_INTERIOR_1` — an **MM** cross-game entrance.
- OoT's `gSaveContext.entranceIndex` is a **direct linear index** into `gEntranceTable[ENTR_MAX = 0x614]`, so applying an MM value reads ~32× past the table and faults at `z_demo.c:2255` (reached via `OoT_Play_Init → Cutscene_HandleConditionalTriggers`). That branch is gated only by an event flag that's unset on a normal save, so it's reachable by a child in the Market — matching the crash log exactly.

## Root cause

The cross-game "startup entrance" is a single **game-agnostic** global. On an OoT→MM switch, `main.cpp` writes the MM target (`0xC010`) into it; the production resume paths (`OoT_Game_Resume` / `MM_Game_Resume`) read it **without clearing**. A stale MM value can therefore survive — e.g. an F10 switch-back to OoT before MM's `Play_Init` consumes it — and OoT applies it to `entranceIndex` with no game check.

MM is structurally immune (its entrance is bit-packed, so an OoT-range value mis-loads a scene rather than reading out of bounds), so the crash and the fix are OoT-focused.

## Changes

- **Game affinity (root cause)** — `src/common/entrance.{cpp,h}`: tag the startup entrance with its destination game and add game-scoped accessors (`*ForGame`) so only the matching game can consume it. `main.cpp` tags the value with the switch's target game; OoT/MM `Play_Init` and `Game_Resume` query `"oot"`/`"mm"`. A leaked MM value is now invisible to OoT, which falls back to its frozen return entrance (the correct resume target). The existing 1-arg setter stays as a wildcard shim, so the dead `switch.cpp` path and existing tests compile unchanged.
- **Defense in depth** — `games/oot/src/code/z_play.c`: bounds-guard the startup entrance in `OoT_Play_Init` (a value `>= ENTR_MAX` is ignored), protecting the crash site and all downstream `gEntranceTable[entranceIndex]` reads in `Play_Init` in one place.
- **Symmetric MM protection** — MM's `Play_Init` and `Game_Resume` use the `"mm"`-scoped accessor so an OoT-tagged value can't drive MM either.
- **Regression test** — `src/common/test_runner.cpp`: `Test_StartupEntrance` now asserts an MM-tagged `0xC010` is invisible to OoT (the exact crash condition), the symmetric case for MM, and that the 1-arg wildcard stays visible to both games. Runs ROM-less in the Tier-1 `redship` ctest suite.

## Verification

- The root cause, the full set of `gEntranceTable` OOB sites, and the fix design were cross-checked with a multi-agent investigation plus three adversarial review passes (breaks-switch / misses-OOB / logic-compile), all of which came back clean.
- Syntax-checked the core `entrance.cpp`/`entrance.h`/`game.c` change locally (`g++ -fsyntax-only`, clean) and confirmed the game-scoped symbols resolve in the edited game code.
- **Not run locally:** a full build/ctest — this checkout has uninitialized submodules, no ROMs/`.o2r` assets, and is network-isolated. CI's ROM-less `redship` tier (including the extended `Test_StartupEntrance` via `AllTests`) covers verification here.

## Fix confirmation (post-fix trace)

For the repro (OoT → Happy Mask Shop → F10 back before MM consumes `0xC010`): `OoT_Game_Resume` sees `Combo_HasStartupEntranceForGame("oot") == false` (value is MM-tagged) → falls back to the frozen return entrance `0x01D1` (Market); `Play_Init` reads `0` for OoT → keeps `0x01D1`; `gEntranceTable[0x01D1]` is in range. No OOB, and OoT resumes at the correct Market entrance. Normal OoT↔MM round-trips are unaffected.

https://claude.ai/code/session_016c1SX9RMsYtFu1ZVDyGwbm

---
_Generated by [Claude Code](https://claude.ai/code/session_016c1SX9RMsYtFu1ZVDyGwbm)_

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8336507109.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8336276138.zip)

### DIFF
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -860,9 +860,13 @@ void MM_Game_Resume(void) {
         // return entrance recorded when we last froze MM. Use the explicit
         // Has accessor because entrance 0x0000 is a valid id and treating it
         // as "unset" would silently drop a legitimate restore.
-        bool hasStartup = Combo_HasStartupEntrance();
+        // Query the MM-scoped accessor: an OoT-tagged value that leaked into
+        // the shared startup global stays invisible to MM, so we fall back to
+        // MM's frozen return entrance instead of spawning from an OoT id
+        // (symmetric with OoT_Game_Resume).
+        bool hasStartup = Combo_HasStartupEntranceForGame("mm");
         uint16_t targetEntrance = hasStartup
-            ? Combo_GetStartupEntrance()
+            ? Combo_GetStartupEntranceForGame("mm")
             : Context_GetFrozenReturnEntrance(GAME_MM);
         gSaveContext.save.entrance = targetEntrance;
         fprintf(stderr, "[MM] Resume entrance: 0x%04X (startup=%u)\n",
--- a/games/mm/src/code/z_play.c
+++ b/games/mm/src/code/z_play.c
@@ -53,6 +53,7 @@ extern uint16_t Combo_CheckEntranceSwitch(uint16_t entranceIndex);
 extern bool Combo_IsCrossGameSwitch(void);
 extern uint16_t Combo_GetStartupEntrance(void);
 extern void Combo_ClearStartupEntrance(void);
+extern uint16_t Combo_GetStartupEntranceForGame(const char* gameId);
 
 s32 MM_gDbgCamEnabled = false;
 u8 D_801D0D54 = false;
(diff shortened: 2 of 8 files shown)


spencerduncan:
Closing as part of cleanup. This was parked once we found the reported crash was actually MM's player-actor spawn (fixed in #352, now merged). The change here — OoT startup-entrance game-affinity + a `>= ENTR_MAX` bounds guard — is still valid latent-bug hardening (green CI, adversarially reviewed), just not the fix for the reported crash. Leaving the branch intact so this can be reopened if the systemic GameInteractor / cross-game-state work doesn't already cover it.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #364, opened 2026-07-18
F10 hotkey switch never freezes state or sets a startup entrance — unfrozen-return hazards


spencerduncan:
Surfaced by the 2026-07-17 Market→Clock Tower crash investigation (the crash itself was the VB enum-aliasing bug, fixed separately; this is a distinct hole).

Entrance-based switches go through `Combo_CheckEntranceSwitch` → `Combo_FreezeState`, so the departing game's SaveContext is snapshotted and a game-tagged startup entrance is set. The F10 hotkey path (`rsbs/src/main.cpp` ~469-472) does neither:

- `Context_HasFrozenState(<departing game>)` stays false, so returning to that game later resumes with whatever the unified `gSaveContext` storage happens to contain — the *other game's* live save bytes.
- No startup entrance is set, so the resumed game re-enters wherever its stale state points.

Concrete hazard found while auditing (games/oot/src/overlays/actors/ovl_player_actor/z_player.c:2847, also :2898, :3619): with `gSaveContext` full of MM residue, `equips.buttonItems`/`heldItemId` hold garbage ids that pass `OoT_Player_ItemToItemAction`'s narrow guard, and `this->upperActionFunc(this, play)` is called unchecked every Player update frame — a wild/NULL indirect call.

Suggested fix: make the F10 path freeze the departing game and tag a return entrance exactly like the entrance path (or refuse to switch when it can't), so "unfrozen resume" becomes unreachable.

spencerduncan:
Re-verified against the phase-2 branch. This is still open, and its failure mode has **changed for the worse** — it is now silent.

`rsbs/src/main.cpp:683-684` adds a new branch:

```cpp
} else if (!isEntranceSwitch && Context_HasFrozenState(nextGame)) {
    Entrance_SetStartupEntrance(Context_GetFrozenReturnEntrance(nextGame), nextGame);
}
```

So the F10 path now *does* set a startup entrance when frozen state exists — but nothing on that path ever **creates** or **clears** that frozen state. Three things confirmed by grep:

1. **F10 still never freezes.** `Combo_CheckHotSwap` (`games/oot/soh/GameExports_SingleExe.cpp:968`) only calls `Combo_RequestGameSwitch()`.
[...]

PR #368, opened 2026-07-18
Phase 2: first full OoT<->MM gameplay round trip

### LINKED ISSUE #366: clang-format CI job is a no-op: run-clang-format.sh targets a nonexistent soh/ root directory
`run-clang-format.sh` (inherited from upstream SoH) does `find soh -type f ...` from the repo root. This repo has no root `soh/` directory (it lives at `games/oot/soh`), so `find` errors to stderr, xargs formats nothing, the script exits 0, and the workflow's `git diff --exit-code` trivially passes:

```
$ bash run-clang-format.sh; echo rc=$?
find: 'soh': No such file or directory
error: cannot use -i when reading from stdin.
rc=0
```

So the `clang-format` check in `.github/workflows/clang-format.yml` has been green on every PR without checking anything.

Fix: point the script at the real trees per CLAUDE.md (`games/oot` and `games/mm` with their respective `.clang-format` configs, plus `src/common` and `rsbs/`), or replace it with a diff-scoped check so it doesn't require reformatting the whole inherited tree at once. Note `xargs` should get `-r` so an empty file list can never invoke `clang-format -i` on stdin.

spencerduncan:
# Phase 2: first full OoT↔MM gameplay round trip

Brings the single-exe combo from "boots both games" to "plays both games across a
switch". The OoT→MM→OoT round trip now completes with correct 3D rendering, audio,
HUD, camera, and doors on both legs.

## What was broken, and why

Every fault in this branch resolved to one of three root-cause classes. All three
are now closed structurally, not patched pointwise.

### 1. Stub signature drift (the dominant crash class)

`src/common/mm_stubs.c` carried hand-written placeholders for MM functions excluded
from the single-exe link. Several had signatures that did not match the real
declarations, so the MSVC x64 ABI silently mismatched registers:

- `OTRConvertHUDXToScreenX` was stubbed `float(float)` against a real
  `int32_t(int32_t)`. Caller read EAX; stub wrote XMM0. MM's A-button viewport
  received stale register contents and collapsed to a degenerate rect — the
  operator-visible "MM is missing the A button". The three-day-clock scissors were
  silently wrong for the same reason.
- The `Gfx_Draw*_DropShadowOverride` family was stubbed `void` while callers consume
  the return value as the display-list **write pointer**. First HUD-visible MM frame
  wrote a gfx command through garbage — WRITE access violation at `0xA7` in
  `MM_Interface_DrawItemButtons`.
- `AudioEditor_GetOriginalSeq` returned 0 for every seqId, funneling all of MM's
  seq-load-status writes into slot 0.

**Structural fix:** these now live in real TUs (`BenPortHudSingleExe.cpp`,
`CosmeticGfxSingleExe.cpp`, `GameExports_SingleExe.cpp`) compiled against the same
headers the callers use, so a drifted signature is a **compile error** instead of a
silent ABI mismatch.

### 2. Implicit declarations (`C4013`)

MM's vendored `PR/*.h` headers had whole prototype blocks fossilized behind `#if 0`
from the vendoring era. Callers compiled against implicit `int` declarations — fine
for integer ABI, a silent scramble for float/pointer params. This is what broke *all*
MM 3D rendering.

**Structural fix:** `/we4013` promotes implicit declaration to a hard error
(`games/mm/CMakeLists.txt`). The revived prototypes are now compiler-checked.

### 3. Cross-game symbol collisions

Both ports export identically-named (and identically-mangled) symbols. Under
`/FORCE:MULTIPLE` first-wins resolution, one game's implementation served both:

- `FrameInterpolation_*` — MM's draws recorded into OoT's interpolation engine and
  replayed OoT's state. MM's 3D world rendered as flashing garbage triangles while
  2D texrects stayed intact.
- `FlagTable` / `SaveEditorWindow` — identical mangled member names, **different
  layouts** (SoH holds a `std::map` where MM holds a `std::vector`). Implicit member
  functions are COMDATs, so one game's copy-constructor served both games' static
  initializers. Crashed at static init; which game won depended on archive pull order.

**Structural fix:** `MM_` symbol-prefix shim headers pulled in via `gfx.h` (so every
`OPEN_DISPS`/`CLOSE_DISPS` expander sees the rename), plus namespace-splitting MM's
side into `S2H`.

## Also fixed

- **Doors vanished after a switch.** `Actor_SpawnTransitionActors` negates entry ids
  *in place* in the process-cached scene resource; only door Destroy callbacks
  un-negate them. A cross-game switch abandons the PlayState without `Play_Destroy`,
  so the departed scene kept all-negative ids for the process lifetime and re-loaded
  with zero doors. Ids are now normalized at scene bind, restoring the N64 fresh-DMA
  invariant.
- **Splash replayed on every return leg.** `CustomLogoTitle` replaces `state.main` via
  `OnZTitleInit`, which turned the `Title_Main` skip gate into dead code. The
  authoritative skip moved to the end of `Title_Init`, past the hook dispatch.
- **Camera stopped following Link.** Camera constants are re-seeded after a switch.
- **MM audio clipping.** The shared RSP mixer is rebased to MM's DMEM window.
- **Scene-setup OOB.** `sceneSetupIndex` can reach 4..19 when a cutscene index
  survives to a scene load, while `headers` is sized by the scene's real alternate
  count. The unguarded index copy-constructed a `shared_ptr` from garbage — a refcount
  write through a wild pointer, i.e. delayed heap corruption presenting as an
  "eventual" crash. Now bounds-checked with a loud fallback to the base header.
- **Malformed saves were fatal.** A base-less `file1.sav` made every boot die on a
  nlohmann `type_error`. SaveManager now never dies on a malformed save and never
  mints one.
- **Portal relocated** to the Clock Tower door. The arrival id (`0xD800`, South Clock
  Town) must stay distinct from the trigger id (`0xC010`) because MM's own cycle
  resets — Song of Time, save-warp, title attract demo — target `0xD800`; when it was
  also the trigger, every in-game reset spuriously fired a cross-game switch.

## Test & CI

- `int-gameplay-roundtrip` harness: programmatic warp/switch driving with door,
  demo-state, camera, and audio-clipping asserts.
- ROM-free locks for the fixed stub contracts (`cosmetic-gfx-stub`), so the
  regressions above are caught without ROM assets.
- `check-symbol-collisions.sh`: nm-based tripwire on the `soh_*`/`2ship_*` defined-
  symbol intersection.
- `run-clang-format.{sh,ps1}`: the previous script targeted a nonexistent `soh/`
  directory and formatted **nothing** — the gate passed vacuously for the life of
  the repo (#366). It now enforces an explicit allowlist,
  `.github/clang-format-paths.txt`, seeded with the 39 files this work touched.

  A full-tree sweep would reformat ~1050 vendored decomp files (~11.6k lines) —
  unreviewable, and permanent conflict surface against the upstream SoH / 2Ship
  trees. Expanding coverage a directory at a time is #369.

  Both scripts read that one file so they cannot drift apart by hand. Three
  anti-vacuity guards all exit 1: missing list, empty list, or any listed path that
  no longer exists (so coverage cannot silently shrink). Verified by exit code, not
  by the green check.
- `RSBS_AUTO_SWITCH_FRAME` / `RSBS_DUMP_GFX` for unattended switch repro and A/B
  render diffing.

Full forensic writeup: `docs/ci-gameplay-repro-postmortem.md`.

## Review findings — tracked, not blocking

This branch was reviewed before merge. Nothing found is a phase-2 regression that
blocks the milestone, but several real bugs were confirmed and filed. We are
pre-alpha; these are tracked rather than gating.

**Critical (fix next):**
- #370 — `SaveManager` never unlocks `saveMtx` on the exception path, so a
  malformed save deadlocks every subsequent save/load. The file is also truncated
  before the write that fails.
- #371 — `sequenceMap` / `gSequenceMap` are `malloc`'d, so the new bounds guards
  leave uninitialized holes that consumers dereference (and `free()`).
- #375 — `check-symbol-collisions.sh` can never fail: the baseline was never
  committed, so it exits 0 in bootstrap mode. Same vacuous-gate class as #366.
- #364 (updated) — the F10 hot-swap path still never freezes, and frozen state is
  never cleared. This branch changed the symptom from an obvious title-screen boot
  into **silent progress rollback**.

**Also filed:** #372 (`GameInteractor_InvertControl` stub returns the enum ordinal
instead of ±1), #373 (MM consumption point does not neutralize live timers, unlike
its OoT twin), #374 (default and test entrance links collide on `0xC010`), #376 (CI
harness gaps), #377 (`RSBS_AUDIO_PROBE` bypasses the handshake it observes), #378
(custom-sequence guard), #379 (cleanup batch), #380 (`ENTR_MAX` unchecked literal).

**Verified sound during review:** the mixer DMEM rebase is *not* an OoT regression —
`DMEM_BUF_SIZE` grows `0xC00`→`0xC90`, the base drops by exactly `0x90`, the window
top is unchanged at `0xFC0`, and OoT's `0x3C0..0xFC0` stays fully contained.

## Known gaps

- The integration tier requires ROM-extracted assets and cannot run on hosted
  runners; it is `workflow_dispatch`-only pending `RSBS_ROM_RUNNER`.
- Normal process exit still heap-corrupts in the CRT onexit chain
  (`0xC0000374`, reproducible with `--version`). Test modes fast-exit via `_Exit`
  to protect verdicts.

Closes #366.

---

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8448540969.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8448466810.zip)


spencerduncan:
## Linux `rando-gen` hang — diagnosed and fixed (it's #383's `ShipInit` ODR, not a randomizer bug)

Investigated on Linux (the operator's box is Windows-only, so this class was invisible there). Summary: the three `rando-gen` CTests never reach seed generation — they **SIGSEGV during boot**, and the crash *handler* then hangs, so it looks like a timeout.

### What actually happens
All three tests crash in `InitOTRForMMFirstBoot → ShipInit::InitAll()`, before generation starts. gdb (catching the raw SIGSEGV; the app handler is suppressed):

```
#0  0x…  ?? ()                                    <- jumps to a heap address (a garbage functor)
#1  std::function<void()>::operator() (this=…)    /usr/include/c++/13/bits/std_function.h:591
#2  ShipInit::Init (path="*")                      games/oot/soh/ShipInit.hpp:46   entry.initFunc()
#3  ShipInit::InitAll ()                           games/oot/soh/ShipInit.hpp:27
#4  InitOTRImpl (...)                              games/oot/soh/OTRGlobals.cpp:1820
#5  InitOTRForMMFirstBoot (...)                    games/oot/soh/OTRGlobals.cpp:1855
#6  Test_RandoGen ()                               src/common/test_runner.cpp:224
```

The "hang" is the **crash handler**: `Ship::ErrorHandler` → `SDL_ShowSimpleMessageBox` (`CrashHandler.cpp:192`) blocks forever in an X11 `poll()` under headless Xvfb, so `exit(1)` never runs. A crash becomes an infinite hang.

### Root cause = the `ShipInit` ODR (#383)
Both ports declare a global `struct ShipInit` whose inline `GetAll()` returns a `static` map, with **different element types**:
- OoT `games/oot/soh/ShipInit.hpp`: `vector<ShipInitEntry>` = `std::function` **+ `std::source_location`** (~40 B)
- MM `games/mm/2s2h/ShipInit.hpp`: `vector<std::function<void()>>` (~32 B)

`GetAll()` is inline → COMDAT-folded to **one** map. Each port's `RegisterShipInitFunc` then `push_back`s its own element type at a **different stride**, corrupting the shared vector. At the crash the `"*"` vector holds 150 entries: the first ~148 are valid OoT registrars, then a corrupt entry (its functor's invoker points into the heap; `registeredAt._M_impl == 0x0`). Both `RegisterShipInitFunc` ctors — MM's 2-arg and OoT's 3-arg — are present in the binary, confirming both writers hit the one map.

### Why it's Linux-only and looks "new"
The ODR predates this PR (`git diff main...HEAD -- '*rando*'` is empty — correct). What changed is the **link**: `f2e86b1` removed the `mm_stubs.c` stubs for `SavingEnhancements_*` / `PauseOwlWarp_*`, so the real MM TUs `2s2h/Enhancements/Saving/SavingEnhancements.cpp` and `Songs/PauseOwlWarp.cpp` — each carrying a `static RegisterShipInitFunc` — now resolve those symbols and their **static initializers run**, polluting OoT's map. Windows hides it (`/FORCE:MULTIPLE` folds differently and happens not to crash).

### Bisect (`76c030df..HEAD`)
Predicate = "SIGSEGV in `ShipInit::Init` under gdb". Only `ccc790f`/`5cc341d` linked and were **GOOD**; the **entire interior fails to link** (the duplicate-def error `f2e86b1` later fixed); `f2e86b1` is the only linkable **BAD**. So the regression is *bounded to* `(ccc790f, f2e86b1]`, and **`f2e86b1` is the unmask point** — the first commit in the range where the link succeeds with the map-polluting MM registrars present. That's exactly why Linux CI never caught it: it was masked by a hard link failure across the whole interior.

### Fix — branch `claude/linux-rando-test-hangs-fhsdfh`, commit `128afbb` (1 file)
Under `RSBS_SINGLE_EXECUTABLE`, move MM's `ShipInit` / `RegisterShipInitFunc` into `namespace S2H` with `using`-declarations — the same `mm_audio_prefix.h` / S2H convention already used in this repo. MM gets its own map; unqualified MM callers compile unchanged; SoH keeps the upstream names. **Deliberately not** worked around with `--allow-multiple-definition` (that would leave the only strict-linker ODR gate — the one #375 can't yet enforce — broken).

Verified on Linux (HEAD + this fix):
- all 3 `rando-gen` variants pass (`rc=0`);
- **24/24** `redship`-labelled tests green;
- link succeeds, no new duplicate symbols.

Ready to cherry-pick/merge `128afbb` into this branch. Not merged from here — leaving that to you since the PR is mid-review.

### Secondary issue (left alone, worth a follow-up)
The crash handler hanging headless (`SDL_ShowSimpleMessageBox` with no WM) is its own bug: any SIGSEGV under Xvfb becomes an infinite hang instead of a fast failure, which is what made this take a real backtrace to localize. A `getenv`/`isatty` guard around the message box would make future crashes fail loudly in CI.

<sub>🤖 diagnosed with Claude Code</sub>

---
_Generated by [Claude Code](https://claude.ai/code)_

spencerduncan:
## Independent review of `128afbb8` (ShipInit ODR fix)

Reviewed on Windows, which cannot reproduce the Linux crash — so this concentrates on stress-testing the reasoning and on Windows build/link correctness. **Verdict: sound, integrated.** Two things worth recording that the diagnosis did not cover.

### Causal chain — verified, not taken on faith

The claim that `f2e86b1b` is what exposed this holds up. Both TUs that commit un-stubbed carry a file-scope registrar whose static initializer runs at load:

- `games/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp:320` — `static RegisterShipInitFunc initFunc(...)`
- `games/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp:232` — same

Removing the `mm_stubs.c` stubs made those symbols resolve to the real TUs, pulling them into the link and running their initializers. Element types differ as described (`soh/ShipInit.hpp:21` `vector<ShipInitEntry>` vs `2s2h/ShipInit.hpp:13` `vector<function<void()>>`).

Note `MotionBlur.cpp` carries **no** registrar — it was only a duplicate-symbol participant, not a map polluter. So the pollution came specifically from the SavingEnhancements/PauseOwlWarp pair.

### The `using`-declaration risk — checked

Global `using S2H::ShipInit;` would be ambiguous in any TU that also sees OoT's global `ShipInit`. Enumerated every includer of either header across the repo: the split is clean, MM files include `2s2h/ShipInit.hpp`, OoT files include `soh/ShipInit.hpp`, and **no TU includes both**. The Windows build graph agrees with the Linux finding.

The `SHIP_INIT_HPP` → `MM_SHIP_INIT_HPP` guard rename is correct hygiene independent of that: both headers previously shared a guard, so a future dual-include would have silently dropped whichever came second — a nasty failure mode to leave armed.

### `ShipInit::InitAll()` — the answer is different from what the task assumed, and it matters

The review task asked whether MM's `InitAll()` at `BenPort.cpp:866` now reads `S2H`'s separate map and whether that silently drops MM enhancement init. Tracing it:

- **`BenPort.cpp` is excluded from single-exe** — `games/mm/CMakeLists.txt:202`, `list(FILTER ship__ EXCLUDE REGEX "2s2h/BenPort\\.cpp$")`. So `BenPort.cpp:866` never runs in this build configuration at all.
- `InitOTRForMMFirstBoot` is **not** in MM's port layer; it is defined in OoT's `games/oot/soh/OTRGlobals.cpp:1851`, and the `ShipInit::InitAll()` in that file (`:1820`) resolves through `#include "soh/ShipInit.hpp"` (`:137`) — i.e. **OoT's** map.
- MM's `GameExports_SingleExe.cpp` contains no `ShipInit` reference, so nothing in the single-exe MM path calls `S2H::ShipInit::InitAll()`.

So the crash was OoT's `InitAll()` walking OoT's map after MM's registrars had appended at the wrong stride, and the fix is correct: OoT's map is now clean.

**No working MM init is dropped** — MM's `InitAll()` was already unreachable in single-exe. But the fix does make explicit and permanent that MM's `ShipInit`-registered enhancements are **inert** in single-exe builds. That is strictly better than the previous state (garbage entries invoked as `std::function`s), so it is not a regression — but it is a real gap, and it is the same registrar-elision problem tracked in #360 / #361, adjacent to #384. Worth a follow-up so nobody assumes MM enhancements initialize today.

### clang-format

The fix was authored with clang-format 18; CI pins 14 and their output differs. Re-checked `games/mm/2s2h/ShipInit.hpp` against the pinned v14 — **clean, no reformat needed**. Added it to `.github/clang-format-paths.txt` so it is enforced going forward, by the same rule as the rest of the list.

### One correction to the plan around #375

The commit comment notes the collision tripwire "skips weak/COMDAT symbols." That is worth pulling out: **committing #375's baseline would not have caught this bug**, because `GetAll()` is an inline COMDAT, not a strong symbol. Any plan that treats #375 as the systemic answer to the ODR class needs it extended to cover COMDAT folding, or this exact shape stays invisible. Noted on #381.

### Also flagging: the headless crash handler turns every crash into a hang

Independent of this fix — `SDL_ShowSimpleMessageBox` in the crash handler blocks forever in an X11 `poll()` on a headless runner, so **any** SIGSEGV under CI presents as a 180 s timeout rather than a fast, legible failure. That cost real time here: the symptom read as "randomizer hang" when it was a segfault at boot. A `getenv`/`isatty` guard would convert it into an immediate crash with a usable exit code. Filed separately.

### Integration

Fast-forwarded onto `claude/ci-cross-game-crash-repro-f6iu2s` (`128afbb8`), plus a follow-up commit adding the allowlist entry. CI running. Windows build of the single-exe target links clean with the namespace change — no regression from `/FORCE:MULTIPLE` resolution.

PR #368 is **not** merged; still mid-review pending `build-linux`.

Out of scope, still open: the remaining #383 items (`GameInteractor` undersized allocation, `UIWidgets::WidgetOptions`) and the #384 registrar collisions.

PR #401, opened 2026-07-20
Give the non-void stubs in soh/stubs.c real returns and supply __osActiveQueue

### LINKED ISSUE #385: Non-void stubs in soh/stubs.c never return; fault handler dereferences the garbage
Found by an automated ODR sweep run after the phase-2 Linux link failure (#368). The sweep started from a duplicate-symbol claim about `__osGetActiveQueue` and **refuted** the link-error framing — but turned up something worse in the process.

## Non-void stubs that never return a value

`games/oot/soh/stubs.c` has several stubs declared with a return type whose bodies are empty:

| Line | Stub |
|---|---|
| 111 | `s32 OoT_osContStartQuery(OSMesgQueue* mq) { }` |
| 124 | `s32 __osDisableInt(void) { }` |
| ~130 | `OSThread* __osGetActiveQueue(void) { }` |
| 133 | `OSThread* __osGetCurrFaultedThread(void) { }` |

Falling off the end of a non-void function is undefined behavior; in practice the caller reads whatever is in the return register. MSVC does not error on this by default, which is why it has survived.

**This is consumed in the crash handler.** `games/oot/src/code/fault.c:537` and `games/mm/src/boot/fault.c:557` both do:

```c
OSThread* iter = __osGetActiveQueue();
```

...and then walk `iter->tlnext`, dereferencing garbage. So the fault handler — the thing that is supposed to give us a usable diagnostic when something goes wrong — is itself liable to fault, which would corrupt exactly the crash reports this project depends on for its whole debugging loop.

## Why `__osGetActiveQueue` is not a link error

Worth recording so nobody re-derives it. There *is* a second definition at `rsbs/src/libultra/os/getactivequeue.c:12`, and both TUs are nominally in the link. But that object references `__osActiveQueue`, which is defined **only** in `games/oot/src/libultra/os/createthread.c:5` and `games/mm/src/libultra/os/thread.c:5` — and both are excluded unconditionally (`games/oot/CMakeLists.txt:207`, `games/mm/CMakeLists.txt:155-157`). libultraship does not supply it either.

So `__osActiveQueue` has no definition anywhere in the link, which makes `rsbs`'s `getactivequeue.o` effectively unpullable: it defines no other symbol, so lazy extraction never reaches for it, and `soh_port`'s `stubs.o` (dragged in unconditionally via `osMemSize` / `OoT_osDestroyThread`) always wins first. If it *were* pulled, the failure would be `undefined reference to __osActiveQueue` — not a duplicate.

That also explains why it did not appear among the eight Linux `multiple definition` errors in #368, unlike the `osWritebackDCache` group in the same file, whose twin lives in libultraship's `os_cache.cpp` — an object that *is* unconditionally pulled.

## Fix

Two coupled steps, because deleting the stub alone would break the link:

1. **Provide the missing storage.** Add `OSThread* __osActiveQueue = NULL;` (or a proper tail-sentinel init) to `rsbs/src/libultra/os/`, since both games' defining TUs are excluded. Without this, `rsbs/src/libultra/os/getactivequeue.c` and `destroythread.c` are dead weight that would break the link if ever pulled.
2. **Delete the stub** at `games/oot/soh/stubs.c:130-131`, so the real `rsbs` implementation resolves — same treatment as the `osWritebackDCache` / `osInvalICache` / `osInvalDCache` group already removed from this file in #368 (see the note left at `stubs.c:108`).

Then audit the remaining no-return stubs listed above and give each an explicit return, or `void`, or a real implementation. A `-Werror=return-type` on this file would prevent recurrence — worth considering alongside the `/we4013` precedent, which turned implicit-declaration into a hard error for exactly this reason.

spencerduncan:
## What was broken

`games/oot/soh/stubs.c` had **24** functions with a return type and no `return` statement. Falling off the end of a non-void function is UB; in practice the caller reads whatever was left in the return register. MSVC treats C4716/C4715 as warnings in C, and `SUPPRESS_WARNINGS` is ON, so none of it was ever reported. Issue #385 named 4; a sweep of the file found 23 empty bodies plus one non-empty body (`_Printf`) that still fell off the end.

The worst instance is `__osGetActiveQueue`. Both games' crash handlers do:

```c
OSThread* iter = __osGetActiveQueue();
while (iter->priority != -1) { ...; iter = iter->tlnext; }
```

(`games/oot/src/code/fault.c:537`, `games/mm/src/boot/fault.c:557`). So the fault handler — whose entire job is to produce a usable diagnostic once something else has already gone wrong — dereferenced a return register as a thread list and was liable to fault itself.

## Mechanism

A real implementation already exists at `rsbs/src/libultra/os/getactivequeue.c`, but it could not win the symbol. It references `__osActiveQueue`, which **nothing in the link defined**: the two defining TUs (`games/oot/src/libultra/os/createthread.c`, `games/mm/src/libultra/os/thread.c`) are both excluded unconditionally, and libultraship supplies neither. That made `getactivequeue.o` unpullable — it exports no other symbol, so lazy archive extraction never reached for it — while `soh_port`'s `stubs.o` is dragged in unconditionally via `osMemSize` / `OoT_osDestroyThread` and always won.

Not a duplicate-definition bug; a **missing-storage** bug wearing one as a disguise. Confirmed: `__osGetActiveQueue` now has exactly one definition in the tree, and `libultraship` defines none of `__osActiveQueue` / `__osRunningThread` / `__osGetActiveQueue` / `__osDispatchThread`, so the new storage cannot collide.

## Fix

**1. `rsbs/src/libultra/os/threadqueue.c` (new)** defines `__osActiveQueue` and `__osRunningThread`.

`__osActiveQueue` points at a tail sentinel with priority `OS_PRIORITY_THREADTAIL` (-1), **not NULL**. The consumers do not null-check, so `NULL` would trade a garbage dereference for a null dereference on the same line of the same handler. An empty list whose head *is* the sentinel makes every walk terminate on the first iteration and answer "no faulted thread" — true for a port that runs no libultra threads. This also makes `destroythread.o` / `getthreadid.o` / `getthreadpri.o` linkable instead of dead weight that would break the link the moment anything reached them.

**2. The stub is deleted**, so the rsbs implementation resolves — the same treatment the `osWritebackDCache` / `osInvalICache` / `osInvalDCache` group got in #368.

**3. The remaining 23 got explicit returns** chosen against each function's real libultra contract, not whatever was cheapest:

| Group | Return | Why |
|---|---|---|
| Controller Pak (`osPfs*`, `__osPfsCheckRamArea`, `osProbeRumblePak`) | `PFS_ERR_NOPACK` | `0` would tell the game a pak exists and send it on to read it. Also preserves the old de-facto behaviour, where non-zero garbage read as "some error". |
| `osContStartQuery`, `osContSetCh`, `osStopTimer`, `osEPiWriteIo` | `0` | Success, per real libultra. `padsetup.c` aborts controller setup on non-zero, so garbage could fail init at random. |
| `osDpGetStatus`, `__osSpGetStatus` | `0` | Idle. Garbage could read as permanently busy. |
| `osSpTaskYielded` | `0` | Not `OS_TASK_YIELDED` — `osSpTaskYield` is a no-op, so re-submitting a task that never yielded is wrong. |
| `osDriveRomInit`, `__osGetCurrFaultedThread` | `NULL` | Documented "no device"/"none"; both fault handlers already branch on NULL. |
| `__osDisableInt` | `0` | Handed straight back to the no-op `__osRestoreInt`. |
| `osGetThreadId` / `osGetThreadPri` | `0` / `OS_PRIORITY_IDLE` | Not the tail priority — callers compare against `-1` to detect end-of-list. |

**4. `_Printf`** — the 24th, which has a body and still fell off the end. `games/oot/src/libultra/libc/sprintf.c` returns `_Printf`'s value verbatim as `sprintf`'s character count, so **every `sprintf()` in the port was handing its caller the return register**. It now returns the number of characters that actually reached the callback (not vsnprintf's would-have-written count, so a truncating format cannot over-report).

**5. Class lockout.** `soh/stubs.c` alone is compiled with `-Werror=return-type` (`/we4716 /we4715` on MSVC), mirroring the `/we4013` precedent that retired the implicit-declaration class. Scoped to this one file deliberately: the vendored decomp tree under `src/` has its own instances and is compiled with `${WARNING_OVERRIDE}` (`/w` or `-w`), so a target-wide escalation would break the build and fight upstream. Source-level `COMPILE_OPTIONS` is emitted after target-level flags, so the escalation wins over `${WARNING_OVERRIDE}` if that ever reaches this file.

## Verification

New ROM-free `redship`-label CTest **`ActiveQueue`** (`--test active-queue`, `src/common/tests/test_active_queue.c`) asserts the three properties `fault.c` actually needs:

1. **Non-NULL** — `iter->priority` is read before any null check.
2. **Bounded termination** — the walk (fault.c's loop verbatim) reaches the `priority == -1` tail within a hard step cap, so a cyclic or garbage list *fails* rather than hanging the tier.
3. **Stability** — two consecutive calls agree; a value read from global storage is invariant, a return register need not be.

Plus `__osGetCurrFaultedThread() == NULL`, `__osDisableInt() == 0`, `OoT_osContStartQuery() == 0`.

Not built locally by design (parallel agent contention) — CI is the build gate. The `-Werror=return-type` escalation is the one thing I could not pre-verify; if it fires, it has found a real 25th instance, which is the point. No file touched is in `.github/clang-format-paths.txt`, so the format gate is unaffected.

## Behavior changes

Real, and all in the safe direction: `sprintf`'s return value is now correct rather than garbage; controller/pak/RSP-status paths now take their deterministic branch instead of a random one. Nothing that previously "worked" depended on a defined value here, because there wasn't one.

## Out of lane

`games/mm/src/code/stubs.c` has the same defect class untouched — `MM___osDisableInt`, `MM_osContStartQuery`, `MM_osGetThreadPri` all have empty non-void bodies. Not in this lane; worth a follow-up issue with the same `-Werror=return-type` treatment.

## Shared-file note

Adds one entry each to `CMake/SingleExecutable.cmake` and `src/common/test_runner.cpp`, which a parallel Wave 2 lane may also be touching. Both edits are additive and adjacent to the `seq-map-bounds` entries, so conflict resolution is trivial.

Closes #385.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8475085518.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8474816957.zip)

### DIFF
--- a/CMake/SingleExecutable.cmake
+++ b/CMake/SingleExecutable.cmake
@@ -298,6 +298,12 @@ if(BUILD_TESTING)
     # bound-arithmetic bugs, so the capacity computation was factored into pure
     # helpers this test calls directly — display-free, ROM-free.
     redship_add_test(NAME SeqMapBounds COMMAND redship --test seq-map-bounds)
+    # Active-thread-queue contract (#385). soh/stubs.c's empty-bodied
+    # __osGetActiveQueue returned the return register, and both games' fault
+    # handlers walk that as a thread list — so the crash handler was itself
+    # liable to crash. Locks non-NULL, bounded termination at the priority == -1
+    # sentinel, and call-to-call stability.
+    redship_add_test(NAME ActiveQueue COMMAND redship --test active-queue)
     # MM cross-game resume contracts (games/mm/2s2h/mm_resume_state_test.cpp):
     # a resume must re-arm the (by-design leaked) system arena for the cold
     # gamestate-chain boot, and Play_Init's startup-entrance consumption must
--- a/games/oot/CMakeLists.txt
+++ b/games/oot/CMakeLists.txt
@@ -134,6 +134,27 @@ list(FILTER soh__Enhancements EXCLUDE REGEX "soh/Enhancements/randomizer/.*")
 # Remove Enhancements from soh__ (will be soh_port lib)
 list(FILTER soh__ EXCLUDE REGEX "soh/Enhancements/.*")
 
+# Falling off the end of a non-void function is a hard error in soh/stubs.c
+# (#385). That file is the port's shim layer: ~90 stubbed libultra entry points,
+# where the failure mode of a missing return is silent — the caller reads the
+# return register. It shipped 24 such bodies. One of them, __osGetActiveQueue,
+# handed a garbage pointer to the crash handler's thread walk
+# (games/oot/src/code/fault.c:537), so the fault handler was itself liable to
+# fault; another, _Printf, was returning the register as sprintf's character
+# count. MSVC does not error on this by default (C4716/C4715 are warnings in C)
+# and SUPPRESS_WARNINGS is ON, which is why it survived to phase 3.
+#
+# Scoped to this one file on purpose: the vendored decomp tree under src/ has
+# its own instances of this pattern and is compiled with ${WARNING_OVERRIDE}
+# (/w or -w), so a target-wide escalation would break the build and would fight
+# upstream. Same shape as the /we4013 precedent that retired the
+# implicit-declaration class in games/mm/CMakeLists.txt.
+if(MSVC)
+    set_source_files_properties(soh/stubs.c PROPERTIES COMPILE_OPTIONS "/we4716;/we4715")
+else()
+    set_source_files_properties(soh/stubs.c PROPERTIES COMPILE_OPTIONS "-Werror=return-type")
+endif()
+
 # Add specific files that don't match the pattern
 list(APPEND soh__Enhancements ${CMAKE_CURRENT_SOURCE_DIR}/soh/Enhancements/savestates_extern.inc)
 list(APPEND soh__Enhancements ${CMAKE_CURRENT_SOURCE_DIR}/soh/Enhancements/speechsynthesizer/DarwinSpeechSynthesizer.mm)
--- a/games/oot/soh/stubs.c
+++ b/games/oot/soh/stubs.c
@@ -36,10 +36,26 @@ f32 qNaN0x10000 = 0x7F810000;
 //	__gSPTextureRectangle(pkt, xl, yl, xh, yh, tile, s, t, dsdx, dtdy);
 // }
 
+/* NOTE (#385): every stub below with a return type returns an explicit value.
+ * Falling off the end of a non-void function is UB — in practice the caller
+ * reads whatever happened to be in the return register — and this file had 23
+ * such bodies. One of them, __osGetActiveQueue, fed a *garbage pointer* to the
+ * crash handler's thread walk (games/oot/src/code/fault.c:537), so the fault
+ * handler was itself liable to fault. soh/stubs.c is compiled with
+ * -Werror=return-type / /we4716 (games/oot/CMakeLists.txt) so the class cannot
+ * come back, mirroring the /we4013 precedent that retired implicit
+ * declarations. Values are the real libultra success/absent codes, never
+ * whatever is cheapest to write. */
+
 OSId OoT_osGetThreadId(OSThread* thread) {
+    /* The port runs no libultra threads; 0 is the id of the (absent) one. */
+    return 0;
 }
 
 OSPri OoT_osGetThreadPri(OSThread* thread) {
+    /* OS_PRIORITY_IDLE. Not the tail priority (-1): callers compare against
+     * that to detect the end of a thread list, never to describe a thread. */
+    return OS_PRIORITY_IDLE;
 }
 
 void OoT_osSetThreadPri(OSThread* thread, OSPri pri) {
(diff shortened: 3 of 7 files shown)


spencerduncan:
## Correction before merge: the `sprintf` impact claim is wrong

The fix is right and should land. One claim in the description is not, and it would otherwise end up in `main`'s history via the squash message.

**Claimed:** *"`_Printf` now returns a correct character count — `games/oot/src/libultra/libc/sprintf.c` returns it verbatim as `sprintf`'s return value, so every `sprintf()` in the port was previously handing its caller the return register."*

**`games/oot/src/libultra/libc/sprintf.c` is commented out in its entirety** — every line, including the `#include`, is prefixed with `//`. There is no `sprintf` in that file to hand anything to anyone. That specific claim is void.

### What the real propagation chain is

Live callers of `_Printf` (word-boundary grep, excluding the definition):

| Site | Uses return value? |
|---|---|
| `src/boot/is_debug.c:22, :32` | discarded |
| `src/code/fault_drawer.c:269` | discarded |
| `src/code/printutils.c:4` | **propagated** — `return _Printf(...)` |

And from there: `printutils.c:4` (`OoT_PrintUtils_VPrintf`) → `gfxprint.c:362` (`GfxPrint_VPrintf`) → and every final consumer (`valueViewer.cpp` and friends) **discards it**.

So the garbage return propagated up two or three frames and was then thrown away by all terminal callers. Real UB, no observed misbehavior.

### Why this still belongs in the PR

Returning nothing from a non-void function is UB, not merely "returns a stale register" — the compiler is free to assume it cannot happen and optimize accordingly, and that assumption gets more dangerous as optimization settings change. Fixing it is correct. The `-Werror=return-type` / `/we4716` escalation is the valuable part, since it retires the class rather than the 24 instances.

**Suggested edit before merge:** replace the `sprintf` sentence with something like *"`_Printf`'s value propagates through `OoT_PrintUtils_VPrintf` → `GfxPrint_VPrintf`, where all terminal callers discard it — real UB with no currently-observable misbehavior."* The other behavior-change notes in the description (Controller Pak `PFS_ERR_NOPACK`, `osContStartQuery`/`osContSetCh`, DP/SP status, `osSpTaskYielded`) I have not re-verified individually and may well be accurate.

Flagging rather than editing the PR body myself so the authoring context stays intact.

PR #423, opened 2026-07-20
TEMP: red-run probe for #401 return-type gate (DO NOT MERGE)

spencerduncan:
Temporary probe PR to empirically verify whether PR #401's per-source
`-Werror=return-type` / `/we4716;/we4715` escalation on
`games/oot/soh/stubs.c` actually fails the build, per the #421 precedent
on the MM side (#392 comment). This PR deliberately empties
`OoT_osSpTaskYielded`'s body (drops `return 0;`).

**This PR is NOT meant to be merged as-is.** Once build-linux/build-windows
results are observed, the breakage will be reverted (if the gate works)
or the gate will be fixed and this replaced with a proper red-then-green
PR (if the gate is decorative).

Tracks: #392, #401, #421.

### DIFF
--- a/games/oot/soh/stubs.c
+++ b/games/oot/soh/stubs.c
@@ -357,11 +357,10 @@ s32 OoT_osStopTimer(OSTimer* timer) {
 }
 
 OSYieldResult OoT_osSpTaskYielded(OSTask* task) {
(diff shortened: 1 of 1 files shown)


spencerduncan:
Verdict: PR #401's -Werror=return-type / /we4716;/we4715 escalation on
games/oot/soh/stubs.c IS live and working — not decorative. Closing
without merging (no code change needed). Full verification details
posted on #392.

PR #425, opened 2026-07-21
TEMP retry: red-run probe for #401 return-type gate (DO NOT MERGE)

spencerduncan:
Retry of #423 — that PR's checks stopped triggering on new pushes for an unknown reason (possible transient GitHub Actions issue). This branch carries the same content (temporary breakage already reverted; HEAD is byte-identical to main plus an empty commit). Supersedes #423, which will be closed. Tracks #392/#401/#421.


spencerduncan:
Diagnostic retry PR — CI runs still aren't triggering here either, so this isn't specific to #423. Closing; superseded by the verdict on #423/#392. No code change needed regardless: the gate already proved itself working via #423's red run (c9d8b959).

Issue #341, opened 2026-07-13
Audit static-initializer registrations in soh_enh / 2ship_* archives for linker elision


spencerduncan:
Side-finding from the PR #340 investigation (song-shuffle seed generation failure).

## Background

In single-exe builds each game is split into static archives (`soh_src`/`soh_port`/`soh_enh`/`soh_rando`, `2ship_*`). Linkers only pull an archive member if it resolves a needed symbol, so any translation unit whose only effect is a `RegisterShipInitFunc` (or similar) static initializer gets **silently dropped** unless it happens to export something referenced — or gets pulled in by COMDAT/inline-symbol luck, which varies by platform and toolchain. Upstream SoH/2Ship never see this because they link all objects directly into an executable.

PR #340 fixed the class for `soh_rando` (WHOLE_ARCHIVE + explicit registrar calls) after confirming `ShuffleSongs.cpp.o` and **all seven `Messages/*.cpp` TUs** were absent from shipped binaries (`nm` ground truth; note `strings`-based checks give false negatives because GCC inlines short `std::string` constructions).

## Remaining exposure

- **`soh_enh`**: dozens of self-registering enhancement TUs (`Enhancements/QoL/*`, `TimeSavers/*`, `cosmetics/*`, etc. — `grep -rn RegisterShipInitFunc games/oot/soh/Enhancements --include=*.cpp | grep -v randomizer` lists them). The ones sampled during the #340 investigation happened to be linked, but each is one refactor away from silently vanishing on one platform and not another.
- **`2ship_enh` / `2ship_rando`**: same pattern on the MM side (2Ship uses equivalent static-registration idioms).

## Suggested approach

Either extend the WHOLE_ARCHIVE treatment to the remaining game archives (verify no duplicate-symbol surprises with the MM_ prefixing and `src/common/mm_stubs.c`, and watch the MSVC 4GB/response-file limits that motivated the split), or add a CI check that compares `_GLOBAL__sub_I_*`/registrar symbols between each archive and the linked binary and fails on drops.

## Verification recipe (from #340)

```sh
# ground truth for a TU's presence in the binary:
nm -C build/redship | grep "RegisterFoo"        # T = linked, absent = dropped
nm -C build/games/oot/libsoh_enh.a | grep ...   # confirm it exists in the archive
```

spencerduncan:
[...]

PR #450, opened 2026-07-21
Fix registrar-elision gate: attribute symbols per-archive, not by basename

spencerduncan:
## Summary

`.github/scripts/check-registrar-elision.sh` (the #341 post-link registrar gate) matched `_GLOBAL__sub_I_<file>` registrar symbols by exact basename against a single deduplicated set of the binary's symbols. Per `docs/unified-surface-findings.md` §4: GCC keys these symbols purely on source **basename**, with no directory/archive component, and gives them **LOCAL** linkage — so two different TUs in different archives sharing a basename produce textually identical symbols that legitimately coexist in the linked binary without colliding. Eleven MM enhancement TU basenames have identically-named twins in OoT's WHOLE_ARCHIVE'd `soh_enh`:

`ArrowCycle, AudioCollection, AudioEditor, EasyFrameAdvance, HyperEnemies, ItemUnequip, MoonJump, N64WeirdFrames, PauseBufferInputs, UnrestrictedItems, WeirdAnimation`

Because `soh_enh` is always fully linked, the old set-membership check always reported MM's copy as "present" through OoT's guaranteed-present twin — whether or not MM's own TU actually survived the link.

## Verification of the artifact (before fixing)

Checked against `origin/main` and a real `build-linux` run (`gh run view 29808211609 --log`, generate-builds on main):

- `lib2ship_enh.a`: 368 registrar symbols, 342 missing → 26 "present" per the old check.
- **None of the 11 colliding basenames ever appear in the script's own MISSING output** — confirmed by grepping the full log for each name. The old check has no way to distinguish "OoT's twin masked MM's elided copy" from "MM's copy is genuinely linked."
- Confirmed via source basename comparison (`games/oot/soh/Enhancements/*` vs `games/mm/2s2h/Enhancements/*`, excluding OoT's `randomizer/` subdir) that exactly these 11 basenames collide — matching the finding exactly.
- Also found one more instance of the same bug class: `Miscellaneous.cpp` collides between `soh_rando`/`2ship_rando` (both **required** archives) — not part of the finding's enh-focused writeup, but the same mechanism, so the fix covers it too.
- The artifact is real; the finding is correct. Not closing out as a false alarm.

## Fix

Replaced set-membership with **occurrence-count attribution**: build a count of each registrar symbol's occurrences in the final binary, then walk archives in their existing declared order (required archives before report-only, unchanged from today) greedily claiming one occurrence per archive's own copy. A colliding basename only credits a later-checked archive once the binary has a genuine occurrence left over after earlier archives already claimed theirs — so `2ship_enh`'s report-only count now reflects what MM actually linked, not what OoT's twin happened to guarantee.

Interface, exit-code contract, and required/report-only distinction are all unchanged. The two Lane C0 reachability probes (`Rando::Spoiler::GenerateFromSaveContext`, `2S2H_RANDO_SPOILER`) are untouched.

**Residual edge case**, documented in the script rather than solved: if two *required* archives share a basename (today just `Miscellaneous.cpp`) and exactly one physical copy is ever dropped, attribution could credit the wrong one of the two — but the overall gate still fails either way, since the shortfall has to land on one of them. It cannot be silently defeated the way the old bug defeated the report-only side.

## Proving the gate still bites (item 4)

Added `--self-test` (mirrors `check-exporter-symbol-collisions.sh --self-test` from #430, for the same "guard the guard" reason), wired into `build-linux` as its own step ahead of the production check. It synthesizes archive/object files with `gcc`/`ar`/`ld -r` (nothing from the real soh/2ship trees) and asserts:

1. Plain no-collision, fully-linked case → PASS.
2. **Required archive's own member dropped from the link (no collision) → gate FAILS.** This is the direct proof that the fix doesn't soften real required-archive elision detection.
3. Basename collision, both copies genuinely linked → both report PRESENT.
4. Basename collision, only the required twin linked → the report-only copy correctly reports **MISSING** (this is the exact bug being fixed — the old logic would have shown it present here).

A local logic-only harness (fake `nm`/counting, no real toolchain) exercised the same six scenarios during development, including the two-required-archives-collision edge case; all passed. Real validation is the CI run on this PR (self-test step runs for real with the container's gcc/ar/ld).

I judged a real red-then-green demonstration commit (temporarily breaking the CMake link setup) disproportionate here since it would touch link topology, which is out of scope for this change — `--self-test` is the mirror of the established #430 precedent instead.

## Corrected numbers (once this PR's CI runs)

Will report the real before/after per-archive counts from this PR's own `build-linux` run in a follow-up comment once CI completes, since that's authoritative ground truth rather than the doc's "roughly" hedge.

## Out of scope note

`games/mm/CMakeLists.txt` has a stale "the two force-linked members, FrameInterpolation and SkipGiantsChamber" comment the findings doc flagged as needing a fix. PR #448 (SavingEnhancements raw-hook migration) is already actively rewriting that exact comment block with real code changes, so I left it untouched here to avoid a merge conflict on contended surface, per this task's coordination note.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8505602169.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8505274711.zip)

### DIFF
--- a/.github/scripts/check-registrar-elision.sh
+++ b/.github/scripts/check-registrar-elision.sh
@@ -10,92 +10,338 @@
 # inlines short std::string constructions, so strings-based checks give false
 # negatives.
 #
+# MEASUREMENT ARTIFACT FIXED HERE (docs/unified-surface-findings.md §4):
+# `_GLOBAL__sub_I_<file>` carries no directory/archive component — it is
+# keyed purely on source basename. These symbols have LOCAL (file-static)
+# linkage, so when two different TUs across DIFFERENT archives share a
+# basename (e.g. games/oot/soh/Enhancements/AudioCollection.cpp and
+# games/mm/2s2h/Enhancements/AudioCollection.cpp — 11 such pairs exist
+# between soh_enh and 2ship_enh, plus Miscellaneous.cpp between soh_rando
+# and 2ship_rando), the linker keeps BOTH as separate, non-colliding symbol
+# table entries when both are linked, but a plain "does this name exist
+# anywhere in the binary" check cannot tell that apart from "only one side's
+# copy survived". Since soh_enh/soh_rando are WHOLE_ARCHIVE'd (always fully
+# linked), their copy is always present — so the old set-membership check
+# always reported the MM twin as "present" too, whether or not MM's own TU
+# actually made it into the link. This inflated 2ship_enh's report-only
+# "present" count (26) with up to 11 basenames that were never really MM's.
+#
+# The fix: track symbol OCCURRENCE COUNTS in the binary, not just presence,
+# and attribute them to archives in the fixed order they are checked below
+# (required archives first). A colliding name is only credited to a
+# later-checked archive once the binary contains MORE occurrences of it than
+# the earlier-checked archives already claimed — i.e. a report-only archive's
+# copy of a colliding basename is "present" only when the binary shows a
+# genuine second occurrence, not merely the required twin's one. This keeps
+# the required-archive check exact (any real drop in a required archive's own
+# registrar count still shows up as a shortfall) while making the
+# report-only numbers for 2ship_enh/2ship_rando_ui trustworthy.
+#
+# One residual edge case, noted rather than solved: if two archives that are
+# BOTH required happen to share a basename (currently just Miscellaneous.cpp,
+# soh_rando/2ship_rando) and exactly one physical copy is ever dropped, the
+# fixed-order attribution may credit the wrong one of the two — but the
+# overall gate still fails either way, since the shortfall has to land on
+# one of them. It cannot silently vanish the way the old bug did.
+#
 # Usage: check-registrar-elision.sh [build-dir]   (default: build-cmake)
+#
+#        check-registrar-elision.sh --self-test
+#            Synthesizes tiny archive/object files with gcc/ar/ld -r (no real
+#            build involved — entirely synthetic, self-contained inputs) and
+#            asserts the counting-based attribution above still (a) passes a
+#            plain no-collision present case, (b) FAILS when a required
+#            archive's own registrar member is dropped from the link, (c)
+#            reports a basename-colliding report-only archive PRESENT when
+#            its own member is genuinely also linked, and (d) reports it
+#            MISSING when only the required archive's same-named member
+#            survived — the exact measurement artifact this fix closes.
+#            Mirrors the check-exporter-symbol-collisions.sh --self-test
+#            added in #430 for the same "guard the guard" reason. Exit 0:
+#            all four scenarios behaved as expected. Exit 1: the detection
+#            logic itself is broken. Exit 2: no usable toolchain found.
 
 set -uo pipefail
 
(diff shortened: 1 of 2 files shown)


spencerduncan:
## Corrected numbers (from this PR's own build-linux run, 29855755804)

Self-test (all 4 scenarios, real gcc/ar/ld in the CI container):
```
self-test: baseline no-collision scenario correctly passed (rc=0, no MISSING)
self-test: required-archive member loss correctly FAILS the gate (rc=1)
self-test: colliding basename with BOTH copies linked correctly reports both PRESENT
self-test: colliding basename with only the required copy linked correctly reports the report-only copy MISSING (rc=0, since report-only never fails the gate)
self-test: OK (all 4 scenarios behaved as expected)
```

Per-archive registrar counts, before (buggy, basename-only) vs after (this fix):

| Archive | Mode | Registrar symbols | Missing (before) | Missing (after) | Present (before) | Present (after) |
|---|---|---|---|---|---|---|
| `libsoh_rando.a` | required | 103 | 0 | 0 | 103 | 103 |
| `libsoh_enh.a` | required | 220 | 0 | 0 | 220 | 220 |
| `lib2ship_rando.a` | required | 161 | 0 | 0 | 161 | 161 |
| `lib2ship_enh.a` | report-only | 368 | 342 | **358** | **26** | **10** |
| `lib2ship_rando_ui.a` | report-only | 3 | 1 | 1 | 2 | 2 |

The required archives are unchanged (as expected — they're all genuinely fully linked; the `Miscellaneous.cpp` collision between `soh_rando`/`2ship_rando` didn't hide anything real, both copies are actually present).

`lib2ship_enh.a`'s report-only count is the one that moved: **26 → 10 present**. Of the 11 basenames that collide with OoT's `soh_enh`, only two are genuinely linked into MM:

- **Genuinely present (7 distinct files, 10 registrar symbols incl. `.cold` variants):** `FrameInterpolation`, `MotionBlur`, `PauseOwlWarp`, `SavingEnhancements`, `SkipGiantsChamber` (never collided, unaffected by the fix), plus `AudioCollection` and `AudioEditor` (collided with OoT twins, but confirmed genuinely linked into MM too — both copies coexist in the binary).
- **Newly correctly reported MISSING** (previously false-"present" via OoT's twin): `ArrowCycle`, `EasyFrameAdvance`, `HyperEnemies`, `ItemUnequip`, `MoonJump`, `N64WeirdFrames`, `PauseBufferInputs`, `UnrestrictedItems`, `WeirdAnimation` — 9 basenames, 16 registrar symbols (incl. `.cold` variants).

This refines the findings doc's "roughly FrameInterpolation, MotionBlur, PauseOwlWarp, SavingEnhancements, SkipGiantsChamber, AudioCollection" — `AudioEditor` also genuinely links (consistent with the `sweet-hopper-1dd88e` branch's "Fix AudioCollection/AudioEditor ODR collision" work, which handled the two together).

Issue #439, opened 2026-07-21
Paired MM world never activates in real gameplay: MM plays vanilla, no foreignPlacements, no MM spoiler


spencerduncan:
Operator playtest (build `33f318b9`, portable Windows zip) reports MM plays **fully vanilla** after generating an OoT randomizer: no MM item shuffle, no foreign OoT items in MM checks. Save-file forensics confirm the pipeline breaks at exactly one seam.

## Forensics (operator's real `.redsave` slots, parsed at pinned ComboContext offsets)

| slot | `sourceIsRando` | `sharedRandoSeed` | `sharedRandoSettingsHash` | `foreignPlacements` | `sharedItemsTagged` |
|---|---|---|---|---|---|
| 0 | 0 | — | `0x00000000` | none | (legacy junk) |
| 1 | 1 | present | `0xde2bd2c6` | **none** | none |
| 2 | 1 | `2108649350` | `0x5dad32ce` | **none** | none |

Slot 2's `sharedRandoSeed` (`2108649350`) is byte-exact the `finalSeed` of the operator's OoT spoiler `Randomizer/32-86-29-43-80.json`.

**Conclusion: Lane B's producer works live** (OoT generation stamps `gComboCtx` correctly through the real GUI flow). **The MM half never ran** — zero `foreignPlacements` in every slot, and **no MM spoiler JSON exists anywhere in the app dir** (only OoT's `Randomizer/*.json`), so `Rando::MiscBehavior::OnFileCreate`'s paired branch did not execute.

## Why CI didn't catch it

`MMRandoGen`'s paired phase drives the OnSaveInit chain **directly**. The natural player flow — enter MM by switching through the Happy Mask Shop (or F10) — goes through the startup-entrance consume path, which may never reach the pairing hook. This is the CI-green/gameplay-broken class of `docs/ci-gameplay-repro-postmortem.md`.

## Hypotheses (root-cause before fixing)

1. **Entry-path gap** — pairing only hooks MM file-select creation; switch-entry never triggers it. *(leading)*
2. **Dispatch-site gap** — the real single-exe `Sram_InitSave` call site doesn't reach the `OnSaveInit` executor, or `IS_RANDO`/saveType gating disarms COND hooks before options are applied.
3. **Harness divergence** — the test's init order/CVar/SHIP_HOME state differs from real `MM_Game_Init`.

## Second bug in the same area

[...]

PR #459, opened 2026-07-21
[DO NOT MERGE] Negative check: does MMPairSwitchEntry catch #439?

spencerduncan:
Throwaway verification branch for #447. Removes only the switch-entry pairing dispatch call in `MM_Play_ConsumeStartupEntrance`, keeping the new `MMPairSwitchEntry` lock intact.

Expected: `MMPairSwitchEntry` FAILS with `FAIL(4)` — the direct OnSaveInit chain pairs but switch-entry does not. If it passes, the lock is vacuous and #447 needs a better one.

Will be closed and the branch deleted once observed.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8505944914.zip)

### DIFF
--- a/CMake/SingleExecutable.cmake
+++ b/CMake/SingleExecutable.cmake
@@ -437,6 +437,21 @@ if(BUILD_TESTING)
         TIMEOUT 180
         ENVIRONMENT "SDL_AUDIODRIVER=dummy;RSBS_DISABLE_OTR_INIT=1")
 
+    # #439: the paired MM world must activate on the flow a player actually
+    # takes. MMRandoGen above drives the OnSaveInit chain DIRECTLY, which is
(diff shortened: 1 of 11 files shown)


spencerduncan:
Verification complete — the lock is non-vacuous.

With only the switch-entry pairing dispatch removed:
- `MMPairSwitchEntry` **FAILED** with `FAIL(4): the direct OnSaveInit chain pairs under master seed 2108649350 but SWITCH-ENTRY did not` — the discriminator correctly identified a dispatch gap rather than a fill dead-end.
- `MMRandoGen` **PASSED**, reproducing the exact CI-green/gameplay-broken condition from the #439 playtest.

Closing; branch deleted.

Issue #460, opened 2026-07-21
Netplay increment 1a: make SharedItem grant semantics network-safe (transport-free, CI-lockable)


spencerduncan:
Tracking issue for **phase 1a** of the netplay increment-1 spike merged in #444.

Design: [`docs/netplay-increment-1-spike.md`](https://github.com/spencerduncan/redshipblueship/blob/main/docs/netplay-increment-1-spike.md) (Status: Proposed).

## Why this is worth doing before any transport decision

The spike's conclusion was that the valuable part of increment 1 is **not the wire**. Four local semantic gaps in the `SharedItem` machinery are invisible today only because the sole producer is an in-process cross-game hand-off. Every one of them is hit by the first remote grant, and every one is fixable and lockable **ROM-free** in the existing `--test` harness.

They are also prerequisites under *every* transport option, so this work cannot be invalidated by the (deferred) 1b transport decision.

## The four gaps

1. **Redemption never fires without a game switch.** `Combo_RedeemSharedItemsForGame` is called only from each game's presence-gated arrival point (`OoT_ConsumeSharedItems` → `games/oot/soh/GameExports_SingleExe.cpp:1055`; `MM_ConsumeSharedItems` → `games/mm/2s2h/GameExports_SingleExe.cpp:1717`). A remote grant for the game you are *already in* would sit un-redeemed until you switched away and back. Needs a third trigger: a gameplay-gated redemption tick calling the same function with the same award callback. No new give path — only a new *when*.

2. **Content de-dup is wrong for a network feed.** `src/common/shared_items.c:56` returns an existing slot unchanged on an `(originGame, id)` match among un-redeemed entries. Correct for a re-fired in-process producer (and documented as such), but it silently **merges two legitimately distinct grants** — two peers sending you the same item before you redeem yields one item. A retransmit and a genuine second gift are indistinguishable at this API. Needs a `grantId` dedup layer *above* a record path that does not apply content de-dup.

3. **Nowhere to put a `grantId`.** `SharedItem` is 4 bytes (`originGame`, `flags`, `id`) with size *and* every member offset static-asserted as `.redsave` format (`src/common/context.h:182-186`). Options and their costs are laid out in the spike §3.

[...]

Issue #440, opened 2026-07-21
Stale cross-game state survives soft reset and new game: old MM clock/flags leak into a new seed


spencerduncan:
Operator playtest (build `33f318b9`): played a rando session on OoT save slot 3, then **soft reset** redship (no process restart), started a **new** rando on save 1, and entered the Happy Mask Shop. MM came up carrying the *previous session's* state — in-game clock where the old session left it, Laundry Pool stray fairy still collected. "The gamestate didn't reset at all."

## Diagnosis

Cross-game session state is **process-global and nothing invalidates it** on soft reset or new-game:

- `gFrozenStates` blobs (`Context_FreezeState`) survive reset,
- the OoT/MM shadow copies survive,
- `gComboCtx` session fields (incl. `sharedItemsTagged`, `foreignPlacements`) survive.

PR #400 made `Combo_ConsumeFrozenState` retire-on-consume, but that only prevents *double* consumption — a stale blob from a dead session is still sitting there for the **first** consume of the next session, which is exactly what the operator hit.

This is also a **cross-seed contamination** bug: a new seed's world can inherit the previous seed's crossings/placements.

Related prior art: the single-exe resume-contract audit (PreNMI `resetTimer` latch, once-only init guards, post-restore boot-chain wipes) — reset semantics need explicit inverses, and the freeze machinery never got one.

## Contract to implement

- **New OoT game (file creation)** invalidates stale cross-game session state: frozen blobs both directions, both shadows, and the previous session's `sharedItemsTagged`/`foreignPlacements`. Seed fields are re-stamped by generation; document each field's fate.
- **Soft reset to title** at minimum invalidates frozen blobs (they describe a session that no longer exists).
- **Loading an existing slot** restores that slot's `.redsave` state; in-memory globals must not shadow it.
- The `.redsave` is per-OoT-slot (`redship_slot{0,1,2}.redsave`) while the in-memory globals are singletons — define clear-and-reload semantics and state explicitly what happens on: reset→same slot, reset→different slot, reset→new file.

## Lock

Headless redship-tier test: freeze state as "session A", simulate the reset/new-game transition, enter MM, assert fresh state. `test_hotswap_freeze.c` already drives the consume machinery.

## Note

The operator's save-1 session data is contaminated by this bug and should be discarded for testing purposes.

spencerduncan:
[...]

PR #463, opened 2026-07-21
Netplay 1a: per-source grant cursors make retransmit vs second-gift decidable; loud overflow; safe-point redemption (ADR 0005, #460)

spencerduncan:
Lands phase **1a** of the netplay increment-1 spike (#444) tracked in #460: the four local semantic gaps in the `SharedItem` grant machinery, fixed **transport-free** and fully CI-locked. No socket, no `BUILD_REMOTE_CONTROL`, no transport — this is the local model plus the producer seam the 1b transport agent writes against. Design record: `docs/adr/0005-sourced-grant-cursors.md`.

## How each gap becomes decidable

| Spike gap | Now |
|---|---|
| Content de-dup merges two peers' identical gifts; retransmit vs second gift indistinguishable | **Per-source monotonic cursor** (`Combo_SubmitSourcedGrant`): `seq <= cursor` is `DUPLICATE` (dropped), `cursor+1` records **without** content de-dup (two gifts = two items), skips are `GAP`. Sourced entries carry `RSBS_SHARED_ITEM_SOURCED` so the in-process content de-dup domain and the sourced domain never mix. |
| No room for a grant id in the pinned 4-byte `SharedItem` | Identity lives in the stream position, not the entry: one `{sourceKey, lastSeq}` cursor per source, carved from `reserved[]` under the ADR 0002 growth contract and serialized — a reload never re-opens the duplicate window. Hosts Archipelago's server-authoritative `ReceivedItems` index directly (index `i` = seq `i+1`; `Combo_GetGrantCursor` is the Sync resume point). |
| Redemption only fires on a game switch | Safe points are defined and locked: the consumer has zero switch-machinery dependency, awards in **received order**, and is single-use under any interleaving. The gameplay-gated tick is specified in the ADR and gets wired by 1b with the first producer that can target the active game (today no producer can, so wiring it would be ROM-only dead code). |
| 64-slot array drops silently on overflow | A full array first **reclaims the oldest redeemed entry** (compaction preserves received order; un-redeemed entries are never dropped). If every slot is un-redeemed: loud refusal — durable serialized `sharedItemOverflowCount` — and for a sourced grant it is **backpressure, not loss**: the cursor does not advance, so the retry is accepted rather than deduped. |

## Save format

Rides the ADR 0002 growth contract: `grantCursors[8]` (offset 672) + `sharedItemOverflowCount` (736) carved from the front of `reserved[]` (332 → 264), offsets pinned by the static-assert chain, zero == unset, `RSBS_SAVE_VERSION` stays 2, `sizeof(ComboContext)` stays 1004. Pre-netplay `.redsave` files keep loading — locked by a crafted legacy-record test.

## #440 composition (not a fix for it)

Cursors live in the same Tier-1 record as the item array on purpose: `ComboContext_Init` retires items + cursors + overflow count atomically, so a dead room's retransmit into a fresh world is a `GAP`, never a silent acceptance — locked in `Test_GrantPersistence` part (c). #440's own fix (making reset actually invoke the invalidation) stays with its owner.

## Locks (all ROM-free, `redship` tier)

- `GrantIdempotency` — retransmit delivers once; second gift (other source, or same source fresh seq) delivers twice; GAP holds everything; source-table exhaustion explicit; in-process content de-dup unchanged and disjoint.
- `GrantRedeemNoSwitch` — received-order awards across interleaved sources with zero switch machinery; single-use across repeated safe points.
- `GrantOverflow` — backpressure with unmoved cursor; refused seq accepted on retry after capacity returns (the no-silent-loss property); durable overflow count for in-process losses; order survives reclamation.
- `GrantPersistence` — cursors/flags/count byte-exact through Save/Load; duplicate window stays closed post-load; pre-netplay legacy record loads with new fields unset; reset retires items + cursors atomically.

Closes nothing yet — #460 stays open for the 1b transport follow-up.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8507293330.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8506960497.zip)

### DIFF
--- a/CMake/SingleExecutable.cmake
+++ b/CMake/SingleExecutable.cmake
@@ -288,6 +288,13 @@ if(BUILD_TESTING)
     # shared structure, the crossing awards exactly once on the OoT side, and
     # gComboCtx.foreignPlacements round-trips through the .redsave record.
     redship_add_test(NAME ForeignItemGive COMMAND redship --test foreign-item-give)
+    # Netplay 1a (ADR 0005, #460): sourced-grant model locks — cursor
+    # idempotency, switch-free received-order redemption, loud overflow with
+    # redeemed-slot reclamation, and .redsave durability + reset atomicity.
+    redship_add_test(NAME GrantIdempotency COMMAND redship --test grant-idempotency)
+    redship_add_test(NAME GrantRedeemNoSwitch COMMAND redship --test grant-redeem-no-switch)
+    redship_add_test(NAME GrantOverflow COMMAND redship --test grant-overflow)
+    redship_add_test(NAME GrantPersistence COMMAND redship --test grant-persistence)
     redship_add_test(NAME ArchiveHotswapLogic COMMAND redship --test archive-hotswap-logic)
     # Unified save (.redsave) headless tests — Phase 2 T6 (#35)
     redship_add_test(NAME SaveRoundtripTiers COMMAND redship --test save-roundtrip-tiers)
--- a/docs/adr/0005-sourced-grant-cursors.md
+++ b/docs/adr/0005-sourced-grant-cursors.md
@@ -88,8 +88,21 @@ source; it is a bug surfaced early.
 
 `RSBS_GRANT_SOURCE_CAP` is 8: an AP server is *one* source regardless of room
 size, P2P co-op is one per peer. Exhaustion is explicit
-(`RSBS_GRANT_NO_SOURCE_SLOT`), never silent. Growing the table later is a
-legal reserved[] carve.
+(`RSBS_GRANT_NO_SOURCE_SLOT`), never silent.
+
+More cursor capacity later is reachable, but **not by bumping
+`RSBS_GRANT_SOURCE_CAP`**. `sharedItemOverflowCount` sits immediately after
+`grantCursors` at offset 736 — a shipped offset the moment this lands — so
+widening the array in place shifts it (and every field carved after it) and
+silently reinterprets every existing `.redsave`. That is a format break, not a
+carve. The legal move is to carve a *second*, separate cursor block from the
+front of the remaining `reserved[]` and have the lookup consult both; each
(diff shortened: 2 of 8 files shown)


spencerduncan:
**Carve collision resolved in your favour — no action needed from you.**

The 1b transport branch had carved `netplayApAppliedCount` + `netplayApRoomFingerprint` from the front of `reserved[]` (332 → 324), immediately after `foreignPlacements` — **byte-for-byte the same region** as your `grantCursors[8]` (672) + `sharedItemOverflowCount` (736). Only one could land.

I discarded mine. Your per-source cursor is the better answer and was designed for exactly this: index `i` = seq `i+1`, `Combo_GetGrantCursor` as the `Sync` resume point. A separate AP-specific cursor would have been a redundant second answer to a question you'd already answered, plus a second path into `Combo_RecordSharedItem` — which #460 forbids. **`reserved[]` is all yours; ADR 0006 makes no carve and no `.redsave` change.**

Worth recording that your seam design survived contact: I audited it against the real Archipelago protocol while measuring 1b, and the `ReceivedItems` index maps onto your cursor exactly as your PR body claims. The backpressure property (refused grant leaves the cursor unmoved, so the retry is accepted rather than deduped) is also precisely what a bounded network inbox needs — it composes with a transport-side bounded queue without either layer needing to know about the other.

Context for 1b: ADR 0006 (#466, merged) measured the Archipelago transport and it is **deferred** — AP validates `Connect` against a multiworld generated from apworlds, so it cannot route us an item until an RSBS apworld exists. Your 1a work is unaffected by that and remains correct under every transport option, which is what the spike predicted.

spencerduncan:
Superseded by #473 — same work rebased onto current main after a power failure interrupted this branch. #473 additionally carries the GrantPersistence fix and the RSBS_GRANT_SOURCE_CAP format-landmine audit commit.

PR #476, opened 2026-07-21
Netplay relay client behind RSBS_NETPLAY (default OFF) — ADR 0007

spencerduncan:
Implements ADR 0007 (merged, `fd2d5802`). **Stacked on #473** — based on `claude/netplay-grant-foundation-2` because `Combo_SubmitSourcedGrant` is the seam and it is not on main yet. Retarget to main and rebase once #473 lands. Draft until then.

## Shape

Three layers, all under `src/common/netplay/` plus one test:

- **`relay_protocol.{h,c}`** — pure codec, no sockets. Byte-at-a-time LE so a big-endian host emits identical bytes with no packing pragma. Decodes the `OP_NONE` keepalive as a NOP: the real server emits one every few idle ticks, so treating a stray `0x00` as a bad opcode would drop a healthy connection.
- **`relay_client.{h,c}`** — state machine feeding `Combo_SubmitSourcedGrant`, the only netplay path into the grant model. No second route into `Combo_RecordSharedItem`.
- **`tests/test_netplay_relay.c`** — five loopback locks.

## ADR 0006 §5's salvage list, carried structurally

Each closes a *verified* Anchor hazard:

- **No receive thread** — the socket is serviced only inside `Relay_Tick()` on the game thread, so hazards #1/#2 cannot exist by construction and ADR 0005's "game thread only" contract needs no marshalling layer.
- **Bounded inbox, recoverable drop** — overflow drops *without* advancing the cursor and flags a replay. Composes with `RSBS_GRANT_RETRY_FULL`: both mean "the source still owes this".
- **Explicit suspend latch** — stops applying, keeps polling. Release is a separate named action.
- **Room binding** — free, because `sourceKey = fnv1a32(roomUuid)` *is* the fingerprint. A different room is a different ADR 0005 source, so a cursor can never be applied against a room that did not produce it.

**Carves nothing from `reserved[]`** — no coordination needed with #473's carve.

## Two bugs a convenient mock would have hidden

Worth calling out because they are exactly the failure class ADR 0006 warned about:

1. The 20-byte join was sent inline right after consuming the server hello. If that send blocked, the join was **dropped on the floor** — the hello was already consumed, so nothing would ever re-send it. Now there is a `JOINING` state meaning "hello done, join still owed".
2. `ChannelSendAll` restarted from zero after a partial send, which would put a prefix **on the wire twice** and desync the server's length-prefixed parser. Now both the join and the outbox track a caller-owned offset.

A never-blocking loopback surfaces neither. They were found by reading, not by the harness — which is itself a data point about §8's honesty.

## What the harness proves, and what it does not

The mock ledger implements multi-server's semantics: append-only, `u64` key dedup, `ledgerBase` replay, and **broadcast to every client including the sender**. That last clause is load-bearing — the real server does *not* skip the sender on `OP_TRANSFER` (only the chat path does), so self-echo is a live condition the client must filter by `targetSlot`. A mock that skipped the sender would hide a real bug, which is the shortcut ADR 0006 §2(b) caught in the AP mock.

**Proves:** retransmit → exactly one item; two peers' same item → two items (the ADR 0005 §1 regression); received-order redemption; grants survive a cross-game switch; late-join catch-up from base 0; rejoin replays as `DUPLICATE` not re-award; foreign payloads skip without desyncing; `RETRY_FULL` leaves the cursor unmoved and the same seq is later *accepted*; overflow is loud; suspend latch; golden byte vectors.

**Does not prove:** that our framing matches the real binary (the mock is a re-derivation from reading `client.c`/`ledger.c`); real socket behaviour (partial reads, `EAGAIN`, mid-frame disconnect, TCP segmentation); that upstream still behaves as read.

Unlike ADR 0006's gap this is a **defect class, not a wall** — multi-server's ledger never inspects the payload, so there is no admissibility check we can fail, and a framing bug is ours to fix unilaterally.

## Default-off, verified in both directions

`RSBS_NETPLAY` defaults OFF, and OFF means *absent*: no netplay source is appended to any target, so the default build gains no TU and links no symbol.

The relay gets **its own CI job** rather than a flag on `build-linux` — deliberately, because `build-linux` packages and uploads the released appimage, and default-off has to mean **shipped-off**. The job asserts both directions: the default configure references no netplay source, and the ON build actually links `Relay_Tick` (so the flag silently ceasing to add sources cannot go green vacuously).

No submodule, no transport library, no `.redsave` change, no auto-connect, no outbound connection in any default path.

## Not in this PR

The **redemption tick** ADR 0005 §4 defined-but-deferred needs a per-frame hook in each game — `games/oot/**` and `games/mm/**`, which is both outside this lane's agreed surface and live under PR #472. Filed as follow-up. Until it lands, relayed grants redeem at the two wired arrival safe points, which is correct but means a grant for the game you are *already in* waits for a switch.

The **real socket layer** is likewise deferred: it is untestable in CI (no network), so landing it alongside a harness that cannot exercise it would be the same category error ADR 0006 called out.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8509429712.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8508652673.zip)

### DIFF
--- a/.github/workflows/generate-builds.yml
+++ b/.github/workflows/generate-builds.yml
@@ -350,6 +350,107 @@ jobs:
           rsbs.appimage
           readme.txt
 
+  # ==========================================================================
+  # Netplay grant relay (ADR 0007, #460)
+  #
+  # A SEPARATE job rather than a flag on build-linux, deliberately: build-linux
+  # packages and uploads the released appimage, so enabling the relay there
+  # would SHIP a feature that is supposed to be off by default. Default-off has
+  # to mean shipped-off, which means the relay never enters a release artifact.
+  #
+  # This job does two things build-linux cannot:
+  #   1. builds with RSBS_NETPLAY=ON and runs the Relay* locks, which do not
+  #      exist in a default configure;
+  #   2. asserts the DEFAULT configure pulls in no netplay translation unit —
+  #      the verifiable form of "the default build is unaffected".
+  # It uploads nothing.
+  # ==========================================================================
+  netplay-relay:
+    needs: generate-rsbs-otr
+    runs-on: ubuntu-22.04
+    container:
+      image: ghcr.io/spencerduncan/redshipblueship-build:latest
+    steps:
+    - name: Git Checkout
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+        fetch-depth: 1
+        fetch-tags: true
+    - name: Configure git safe directory
+      run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+    - name: Install packaging tools
+      run: |
+        apt-get update
+        apt-get install -y imagemagick file desktop-file-utils
+    - name: Configure ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        save: true
+        max-size: "1.5G"
+        key: ${{ runner.os }}-ccache-netplay-${{ github.ref }}
+        restore-keys: |
+          ${{ runner.os }}-ccache-netplay-${{ github.ref }}
+          ${{ runner.os }}-ccache-${{ github.ref }}
+          ${{ runner.os }}-ccache-refs/heads/main
+          ${{ runner.os }}-ccache
+    - name: Download soh.o2r
+      uses: actions/download-artifact@v4
+      with:
+        name: soh.o2r
+        path: build-cmake/soh
+    - name: Download 2ship.o2r
+      uses: actions/download-artifact@v4
+      with:
+        name: 2ship.o2r
+        path: build-cmake/mm
+    - name: Assert the DEFAULT build pulls in no netplay TU
+      # Configure-only (seconds, no compile) in the default configuration and
+      # grep the generated build system. With RSBS_NETPLAY=OFF the netplay
+      # sources are never appended to REDSHIP_COMMON_SOURCES, so no rule
+      # referencing them can exist. This is the negative control for the
+      # positive one below; without it "default OFF" is a claim about CMake
+      # that nothing checks.
+      run: |
+        cmake --no-warn-unused-cli -H. -Bbuild-default -GNinja -DCMAKE_BUILD_TYPE:STRING=Release >/dev/null
+        if grep -rq "netplay/relay_" build-default/build.ninja; then
+          echo "FAIL: default configure references a netplay source; RSBS_NETPLAY=OFF must mean absent."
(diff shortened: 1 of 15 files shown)


spencerduncan:
## Fully green, gated only on #473

All checks pass on the stacked base — including the new `netplay-relay` job, which is the point of this PR:

```
OK: default configure pulls in no netplay translation unit.
OK: relay symbols linked.
1/5 Test #20: RelayWireFormat ....... Passed
2/5 Test #21: RelayLoopback ......... Passed
3/5 Test #22: RelayCatchup .......... Passed
4/5 Test #23: RelayBackpressure ..... Passed
5/5 Test #24: RelaySuspendLatch ..... Passed
100% tests passed, 0 tests failed out of 5
...
100% tests passed, 0 tests failed out of 53   # full ROM-free tier, relay built
```

Both controls fire: the negative one (a default configure references no netplay source) and the positive one (`Relay_Tick` is actually in the ON binary, so the flag silently ceasing to add sources cannot pass vacuously). The 53-test tier confirms the relay perturbs nothing else.

`build-linux` 4m22s, `build-windows` 32m28s, `clang-format` and `clang-tidy` green.

## Landing plan

**#473 is itself green, CLEAN and mergeable** — it is the only thing in the way. I deliberately did not merge it: it is not my PR, and merging another author's work is not mine to do.

Once #473 lands, this does **not** want a rebase — #473 squash-merges to a single commit, so replaying its individual commits would conflict with itself. The clean move is a fresh branch off `main` with **this branch's one commit cherry-picked**, then retarget to `main` and undraft. The commit was authored directly on top of #473's tree, so it should apply as-is.

## Scope note

Two things are deliberately **not** here, and both are in the PR body:

- The **redemption tick** ADR 0005 §4 defined-but-deferred needs a per-frame hook in each game (`games/oot/**`, `games/mm/**`) — outside this lane's agreed surface and live under #472. Until it lands, relayed grants redeem at the two wired arrival safe points, so a grant for the game you are *already in* waits for a switch.
- The **real socket layer**. It is untestable in CI (no network), and landing it beside a harness that cannot exercise it would repeat the category error ADR 0006 named.

spencerduncan:
Superseded by the restacked PR — same commit cherry-picked onto main now that #473 has squash-merged, per the cherry-pick-not-rebase note above.

Issue #516, opened 2026-07-24
Systemic: BenPort.cpp's exclusion elides InitOTR's whole registration list — CustomItem, CustomMessage and GfxPatcher registrars are absent from the binary


spencerduncan:
## Summary

`games/mm/2s2h/BenPort.cpp` is excluded from the single-exe build (`games/mm/CMakeLists.txt:238`). It is the **sole caller** of a whole set of `Init*` / `Register*` functions, and nothing in the single-exe path replicates that call list. `2ship_enh` is a plain `STATIC` archive, so those TUs are then link-elided entirely.

Net effect measured from `build-cmake/redship.map`: **only 8 of ~130 `2ship_enh` translation units survive the link.**

This is the same class as #513 (found first, and narrower than it appeared). #513 should be treated as one instance of this issue.

## Confirmed absent from the binary

Symbol census against `build-cmake/redship.map` (MSVC Release). `0` = not in the binary:

| Symbol | Map hits | Sole caller |
|---|---|---|
| `CustomItem::RegisterHooks` | **0** | `BenPort.cpp:872` |
| `CustomItem00_Init` (whole TU elided) | **0** | — |
| `CustomMessage::RegisterHooks` | **0** | `BenPort.cpp:873` |
| `GfxPatcher_ApplyNecessaryAuthenticPatches` | **0** | `BenPort.cpp:869` |
| `InitEnhancements` | **0** | `BenPort.cpp:867` |
| `RegisterSavingEnhancements` | **0** | `Enhancements.cpp:5` (via `InitEnhancements`) |

`RegisterAutosave` is a subtlety worth stating precisely: its *definition* is present (SavingEnhancements.cpp is one of the 8 surviving TUs), but its only caller is `InitEnhancements()`, which is absent — so it is **present-but-never-called**. Presence in the map is not reachability.

## Player impact — two of these are critical for randomizer

**1. `CustomItem::RegisterHooks` — every randomized custom item is uncollectable.**
`CustomItem::Spawn` creates an `ACTOR_EN_ITEM00` with params `ITEM00_NOTHING` and depends entirely on a `ShouldActorInit` hook to swap in `CustomItem00_Init/Update/Draw/Destroy`. With the registrar elided, that swap never happens and the actor stays a nothing-item. Live on any save with `saveType == SAVETYPE_RANDO`.

Note this became *reachable* only after #512 — before it, `ShouldActorInit` was not dispatched at all, so this was masked by a larger bug.

[...]

PR #519, opened 2026-07-24
Seed an unset lastTimeLog instead of injecting a Unix epoch into filePlaytime (#513)

### LINKED ISSUE #513: MM's first pause/owl save injects a full Unix epoch (~56 years) into filePlaytime: the adder is linked but its baseline seeder is elided
## Summary

In single-exe, MM's first pause/owl save injects roughly a **full Unix epoch (~1.78e9 seconds, ~56 years)** into `shipSaveInfo.filePlaytime`, which then serializes to disk. Found during an audit of the hook-dispatch bug class behind #511.

This is **not** a hook-dispatch bug — it survives fixing those — so it needs its own fix.

## Mechanism

`SavingEnhancements_AdvancePlaytime` (`games/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp:107`) computes a delta against a baseline:

```c
gSaveContext.save.shipSaveInfo.filePlaytime += timestamp - gSaveContext.shipSaveContext.lastTimeLog;   // :110
gSaveContext.shipSaveContext.lastTimeLog = timestamp;                                                  // :111
```

The **only** writer that seeds `lastTimeLog` is inside a `COND_HOOK(OnSaveLoad, ...)` registered by `RegisterSavingEnhancements()`:

```c
gSaveContext.shipSaveContext.lastTimeLog = GetUnixTimestamp();   // :247
```

And `lastTimeLog` is explicitly zeroed on the save paths at `games/mm/src/code/z_sram_NES.c:1033` and `:1265`.

Meanwhile `SavingEnhancements_AdvancePlaytime` is called from **plain C, not a hook** — so it runs regardless of any registry wiring:

- `games/mm/src/code/z_sram_NES.c:2212`
- `games/mm/src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope_NES.c:3583`

So the adder runs with `lastTimeLog == 0`, and `timestamp - 0` is the whole epoch.

## Why the registrar never runs

`RegisterSavingEnhancements()` is reachable only via `InitEnhancements()` (`2s2h/Enhancements/Enhancements.cpp:5`), which is called only from `2s2h/BenPort.cpp:867` — and BenPort.cpp is excluded from single-exe at `games/mm/CMakeLists.txt:238`. `2ship_enh` is a plain `STATIC` archive (`CMakeLists.txt:372`), so with no inbound reference the TU is elided.

## Decisive evidence — the linked binary, not inference

Symbol census of `build-cmake/redship.map` (MSVC, Release, built at `b737ae49`):

| Symbol | Occurrences in map |
|---|---|
| `SavingEnhancements_AdvancePlaytime` | **2** (present) |
| `RegisterSavingEnhancements` | **0** (absent) |
| `InitEnhancements` | **0** (absent) |

The adder is in the binary; the seeder is not. That is the bug, straight from the link.

## Impact

`filePlaytime` and `fileCompletedAt` both serialize (`BenJsonConversions.hpp:124-125`), so the corrupt value is durable. No current in-game surface reads `filePlaytime` in this build — `DisplayOverlay.cpp:78` does, but that overlay isn't reachable in single-exe yet — so this is silent today and becomes visible the moment MM's overlay/file-select surfaces land, at which point existing saves already carry the bad value.

Compounding: `OnGameCompletion` is also dead (stub at `src/common/mm_stubs.c:161`), so `fileCompletedAt` is never stamped and the `fileCompletedAt == 0` guard at `:108` stays true forever — playtime keeps accumulating after credits too.

## Suggested fix

Two independent options; the first is the minimal one:

1. **Seed defensively at the point of use.** In `SavingEnhancements_AdvancePlaytime`, treat `lastTimeLog == 0` as "first observation": set it to `GetUnixTimestamp()` and add nothing this call. Robust regardless of registrar wiring, and fixes existing behavior immediately.
2. **Give `RegisterSavingEnhancements()` a single-exe home** — e.g. a `RegisterShipInitFunc`, as `SavingEnhancements.cpp:324` already does for `RegisterRememberSaveLocation`. Correct in the larger sense, but re-arms every other hook in that TU at the same time, so it wants its own review.

Doing (1) now and (2) deliberately later seems right.

**Repair for already-written saves:** a `filePlaytime` larger than, say, 10 years is unambiguously this bug and can be clamped to 0 on load.

Refs #511, #438

spencerduncan:
Fixes #513 — MM's first pause/owl save injects a full Unix epoch (~56 years) into the persisted `filePlaytime`. A point-of-use fix that is correct independent of the elided registrar (#516) it originally stemmed from.

## Root cause

`SavingEnhancements_AdvancePlaytime` (`SavingEnhancements.cpp:107`) accrues:

```c
filePlaytime += now - lastTimeLog;
lastTimeLog = now;
```

`filePlaytime` is a **persisted** field (`z64save.h` `ShipSaveInfo`); `lastTimeLog` is **runtime-only** (`ShipSaveContext`, documented "not persisted"). `lastTimeLog` is seeded only by `RegisterSavingEnhancements`' `OnSaveLoad` hook — which the single exe elided entirely (#516; revival is Phase 2) — and is re-zeroed on new-file / continue paths (`z_sram_NES.c:1033/1265`).

But `AdvancePlaytime` is called from **plain C** (`z_sram_NES.c:2212`, `z_kaleido_scope_NES.c:3583`), independent of any hook wiring. So it runs with `lastTimeLog == 0` and writes `now - 0` — a whole Unix epoch (~1.7e9 s) — to disk.

Confirmed from the linked binary in #513: `SavingEnhancements_AdvancePlaytime` is present in `redship.map`, its `RegisterSavingEnhancements` seeder is absent (0 hits). The adder ships; its baseline seeder does not.

## Fix

Treat `lastTimeLog == 0` as the seed: set it, accrue nothing that tick.

```c
if (gSaveContext.shipSaveContext.lastTimeLog != 0) {
    filePlaytime += now - lastTimeLog;
}
lastTimeLog = now;
```

This is deliberately at the point of use, not dependent on the seeder:
- Correct whether or not #516 Phase 2 revives the `OnSaveLoad` seeder.
- Survives the `z_sram_NES.c` re-zeroing, which would otherwise re-arm the bug mid-session even after the seeder lands.
- Costs at most one interval's accrual (file-load → first save point), negligible against injecting 56 years.

## Testing

New ROM-free row **`MMPlaytimeSeed`**, three checks:
1. **The bug** — `lastTimeLog == 0` ⇒ `filePlaytime` stays near zero (pre-fix: ~1.7e9), `lastTimeLog` seeded.
2. **Accrual still works** — `lastTimeLog == 1` (non-zero) ⇒ `filePlaytime` grows, proving the guard narrows only the zero case.
3. **The `fileCompletedAt` gate is intact** — a completed file blocks all accrual.

Falsified before trusting (see PR comment). Local `ctest` green on both `redship` and `rando` tiers.

## Scope note

This prevents **new** corruption. Repairing saves already written with the bad value (a load-time clamp of an absurd `filePlaytime`) is deferred: nothing in the current build reads `filePlaytime` (`DisplayOverlay.cpp:78` isn't reachable in single-exe yet), so the dormant value causes no active harm, and a clamp touches the deserialization path — better handled when MM's overlay surfaces land and there's a reader to validate against. Noted on #513.

Closes #513
Refs #516

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8600275861.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8599948334.zip)

### DIFF
--- a/.github/clang-format-paths.txt
+++ b/.github/clang-format-paths.txt
@@ -40,6 +40,7 @@ games/mm/2s2h/mm_culling_test.cpp
 games/mm/2s2h/mm_fb_effects_test.cpp
 games/mm/2s2h/mm_flash_filenum_test.cpp
 games/mm/2s2h/mm_hook_dispatch_test.cpp
+games/mm/2s2h/mm_playtime_seed_test.cpp
 games/mm/2s2h/mm_framebuffer_effects.c
 games/mm/2s2h/mm_gi_shim_test.cpp
 games/mm/2s2h/mm_notification_binding_test.cpp
--- a/CMake/SingleExecutable.cmake
+++ b/CMake/SingleExecutable.cmake
@@ -523,6 +523,13 @@ if(BUILD_TESTING)
     # randomized while their models and dialog stayed vanilla. Display-free and
     # ROM-free, so it runs in this redship tier.
     redship_add_test(NAME MMHookDispatch COMMAND redship --test mm-hook-dispatch)
+    # filePlaytime epoch injection (#513): SavingEnhancements_AdvancePlaytime is
+    # called from plain C (z_sram_NES.c, z_kaleido_scope_NES.c) regardless of
+    # hook wiring, and accrued now - lastTimeLog with lastTimeLog unseeded (0) --
+    # the OnSaveLoad seeder was elided (#516) and the value is re-zeroed on
+    # new-file paths -- writing a ~56-year Unix epoch into the PERSISTED
+    # filePlaytime. Display-free and ROM-free, so it runs in this redship tier.
+    redship_add_test(NAME MMPlaytimeSeed COMMAND redship --test mm-playtime-seed)
     # MM tracker registration surface (#392): the four MM tracker windows must
     # register on the shared Gui under "MM "-prefixed names (SoH owns the
     # unprefixed ones; Gui::AddGuiWindow rejects duplicates) and their
--- a/games/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
+++ b/games/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
@@ -107,7 +107,18 @@ extern "C" bool SavingEnhancements_CanSave() {
 extern "C" void SavingEnhancements_AdvancePlaytime() {
     if (gSaveContext.save.shipSaveInfo.fileCompletedAt == 0) {
         uint64_t timestamp = GetUnixTimestamp();
-        gSaveContext.save.shipSaveInfo.filePlaytime += timestamp - gSaveContext.shipSaveContext.lastTimeLog;
+        // lastTimeLog == 0 means "no prior observation this session": either the
+        // OnSaveLoad seeder in RegisterSavingEnhancements has not run (in the
+        // single exe it was elided entirely — #516 Phase 2 revives it), or a
+        // new-file / continue path zeroed it (z_sram_NES.c). filePlaytime is a
+        // PERSISTED field (z64save.h ShipSaveInfo) while lastTimeLog is not
+        // (ShipSaveContext), so accruing `timestamp - 0` here writes a whole
+        // Unix epoch (~1.7e9 s, ~56 years) to disk. Treat the zero as the seed
+        // and accrue from the next tick instead — correct whether or not the
+        // seeder ran, and it survives the z_sram_NES.c re-zeroing (#513).
+        if (gSaveContext.shipSaveContext.lastTimeLog != 0) {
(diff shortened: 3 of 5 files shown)


spencerduncan:
Falsified before trusting: reverting the `lastTimeLog != 0` guard and rebuilding fails `FAIL(1): AdvancePlaytime injected an epoch into filePlaytime from lastTimeLog == 0`; restoring it returns green. Local `ctest` 60/60 on the redship tier, 13/13 on rando.

Issue #438, opened 2026-07-21
Port MM dispatch for the remaining dormant hook types (14 of 23 remain; OnOpenText/OnActorInit/OnSceneInit now dispatch)


spencerduncan:
Follow-up to PR #435 (the #392 VB-dispatch fix), which wired the five "Should" dispatchers (`GameInteractor_Should` VB verdicts + `ShouldActorInit`/`ShouldActorUpdate`/`ShouldActorDraw`/`ShouldItemGive`) through the MM-owned `S2H::GameHooks` registry via the macro rebind at the bottom of MM's `GameInteractor.h`.

The linked MM rando set (2ship_rando) REGISTERS the following hook types that still have **no MM dispatch point** — their registrations park in `S2H::GameHooks` and never run, because the corresponding `GameInteractor_Execute*` names still resolve to OoT's game-gated wrappers or `src/common/mm_stubs.c` no-ops (registration counts from `COND_HOOK`/`COND_ID_HOOK` sites in `games/mm/2s2h/Rando/**`):

| hook type | registrations | current provider | note |
|---|---|---|---|
| `OnOpenText` | 98 | OoT gated wrapper (`(u16*, bool*)`) | the rando hint system + textbox overrides — the largest dormant surface |
| `OnActorInit` | 16 | OoT gated wrapper | actor-behavior arming that isn't ShouldActorInit-based |
| `OnSceneInit` | 4 | OoT gated wrapper — **signature drift**: MM `(s16 sceneId, s8 spawnNum)` vs OoT `(int16_t sceneNum)` | must be typed MM-side, like the OnFlagSet family in PR #428 |
| `OnActorKill` | 2 | none (unresolved via stubs?) / gated | |
| `OnActorDestroy` | 2 | gated wrapper | |
| `OnActorDraw` | 1 | `mm_stubs.c` void no-op | |
| `OnKaleidoUpdate` | 1 | `mm_stubs.c` no-op | |
| `AfterKaleidoDrawPage` | 1 | `mm_stubs.c` no-op (drifted `(void*, int)`) | |
| `BeforeEndOfCycleSave` | 1 | `mm_stubs.c` no-op | three-day-reset check handling |
| `AfterEndOfCycleSave` | 1 | `mm_stubs.c` no-op | |

[...]

PR #521, opened 2026-07-24
Wire OnGameCompletion dispatch, the last revived-registrar gap (#438)

spencerduncan:
Wires `OnGameCompletion` dispatch — the single hook that gained a live registrant from #520 but still bound a `mm_stubs.c` no-op. Same 3-file pattern as #512/#514, verified by a re-scan of every revived registrar (the audit's full re-scan is on [#438](https://github.com/spencerduncan/redshipblueship/issues/438)).

## The gap

After #518/#520 un-elided `CustomItem`/`CustomMessage`/`GfxPatcher`/`RegisterSavingEnhancements`/`RegisterAutosave`, I checked every hook type those five register against the dispatched set. **Exactly one gap:** `RegisterSavingEnhancements` registers `COND_HOOK(OnGameCompletion, …)` (`SavingEnhancements.cpp:268`), but dispatch bound the no-op stub at `mm_stubs.c:161`. Every other hook they touch (`ShouldActorInit`, `OnOpenText`, `OnGameStateUpdate/DrawFinish`, the VB and save hooks) is already dispatched.

The audit originally listed this as "not fixable by dispatch alone — `RegisterSavingEnhancements` has no single-exe caller." #520 revived that caller, so it's now pure dispatch work.

## What it does

The hook stamps `shipSaveInfo.fileCompletedAt` and freezes post-credits playtime when the game is beaten. Its call sites — `z_boss_07.c:1777` (Majora's Wrath death) and `Rando/GiveItem.cpp:118` (final Triforce piece) — reached the stub, so the stamp never happened.

## Changes (the established pattern)

- **`GameExports_SingleExe.cpp`** — `MM_GameHooks_ExecuteOnGameCompletion` calling `S2H::GameHooks::Execute<OnGameCompletion>()`. Single 0-arg leg (`DEFINE_HOOK(OnGameCompletion, ())`, no id/ptr/filter registrants), matching the excluded `GameInteractor.cpp` twin.
- **`GameInteractor.h`** — declaration + `#define` rebind in the single-exe block.
- **`mm_stubs.c`** — delete the no-op, with the file's standard "re-stubbing would sever dispatch" comment. `OnGameCompletion` is MM-only (no OoT twin), so deleting it can't orphan an OoT caller — and it makes a future dispatch miss a *link* error, not a silent no-op.

## Headless-safety

The hook body touches only `gSaveContext` + `GetUnixTimestamp` + `SavingEnhancements_AdvancePlaytime` — no `MM_gPlayState` — so it doesn't reproduce the #516 Phase 1 headless crash and needs no play-state scaffolding. Both call sites are play-state paths that never run in a ROM-free test anyway.

## #513 interaction (resolved)

`AdvancePlaytime` reads `lastTimeLog` as its baseline; that baseline is seeded now (#520 revived the `OnSaveLoad` seeder) and guarded (#519), so the post-credits stamp banks correct playtime, and both `AdvancePlaytime` and the hook body gate on `fileCompletedAt == 0` so accrual stops after the stamp.

## Testing

`MMHookDispatch` gains check 8: registers a throwaway `OnGameCompletion` hook on `S2H::GameHooks`, drives the dispatcher through the upstream spelling, asserts it fired. Because the symbol is MM-only and the stub is deleted, dropping the rebind or dispatcher is a **link error** — check 8 additionally proves the dispatcher reaches the registry at runtime. Local `ctest` green on `redship` and `rando` tiers (falsification in PR comment).

Refs #438, #520, #513

### Build Artifacts
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8604322975.zip)
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8604210349.zip)

### DIFF
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -1923,6 +1923,21 @@ extern "C" void MM_GameHooks_ExecuteAfterEndOfCycleSave(void) {
 }
 
 /**
+ * OnGameCompletion dispatch (#438). The last revived-registrar gap: #520
+ * un-elided RegisterSavingEnhancements, whose COND_HOOK(OnGameCompletion)
+ * (SavingEnhancements.cpp) stamps shipSaveInfo.fileCompletedAt and freezes
+ * post-credits playtime — but the call sites (z_boss_07.c Majora's Wrath death,
+ * Rando/GiveItem.cpp final Triforce) reached the mm_stubs.c no-op, so the stamp
+ * never happened. 0-arg (DEFINE_HOOK(OnGameCompletion, ())), single Execute<>
+ * leg — no MM TU registers it by id/ptr/filter — matching the excluded
+ * GameInteractor.cpp twin. The hook body touches only gSaveContext + playtime,
+ * no MM_gPlayState, so it is headless-safe.
+ */
+extern "C" void MM_GameHooks_ExecuteOnGameCompletion(void) {
+    S2H::GameHooks::Execute<GameInteractor::OnGameCompletion>();
+}
+
+/**
  * MM-owned "Should" dispatch (#392 VB follow-up). MM call sites reach these
  * through the single-exe macro rebind at the bottom of MM's GameInteractor.h
  * (GameInteractor_Should -> MM_GameHooks_ExecuteVBShould, etc.), so MM VB ids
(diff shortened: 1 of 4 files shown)


spencerduncan:
Falsified before trusting: neutering the `MM_GameHooks_ExecuteOnGameCompletion` dispatcher body fails `FAIL(8): OnGameCompletion registrant never ran`; restoring it returns green. (Dropping the rebind or the dispatcher entirely is a *link* error, since the symbol is MM-only and the stub is deleted — so check 8 specifically locks that the dispatcher reaches the registry.) Local ctest green on redship (60) and rando (13).

Issue #530, opened 2026-07-29
Only the owl statue got #527's unified capture — Song of Time new-cycle, pause save-and-quit, game-over save, special saves, and autosave still persist zero bytes in a cross-game MM session while presenting a completed save


spencerduncan:
The MM-side owl capture writes a Tier-1 (`gComboCtx`) snapshot that is *newer* than OoT's own `file{N+1}.sav`. On reload, `#528` restores that newer Tier-1 wholesale while deliberately leaving Tier-2 unarmed, so OoT resumes from its older `.sav` — but the `RSBS_SHARED_ITEM_REDEEMED` records committed in Tier-1 survive. The redeem loop skips REDEEMED entries forever, so any cross-game item delivered to OoT after OoT's last save point is permanently lost.

## Mechanism

Anchors at `f4ad479b`.

1. Pre-#527, every `.redsave` write rode OoT's `OnSaveFile`, so Tier-1 and OoT's world state hit disk in the same instant and could not diverge.
2. `MM_Combo_CaptureSaveToUnifiedSlot` (`games/mm/2s2h/GameExports_SingleExe.cpp:1596-1614`) now serializes `gComboCtx` on the MM owl-save path — a moment at which OoT's `.sav` is *not* updated.
3. `Combo_CheckEntranceSwitch` (`games/oot/soh/GameExports_SingleExe.cpp:1139-1190`) freezes the OoT shadow and publishes the slot on the OoT→MM crossing, but contains no `Save_SaveFile` call. OoT's on-disk `.sav` therefore still reflects the last time the player used an actual OoT save point.
4. Shared items are redeemed on OoT arrival by mutating only the **live** SaveContext and setting `RSBS_SHARED_ITEM_REDEEMED` in `gComboCtx`; the redeem loop skips REDEEMED entries unconditionally on every later pass (`shared_items.c:271-289`).
5. On load, `#528` restores the newer `gComboCtx` wholesale (`save.cpp:313`) while intentionally leaving Tier-2 unarmed — "file{N+1}.sav is the authority" (`save.cpp:326-330`).

Net: the REDEEMED *record* is durable; the item the record accounts for is not.

## Evidence

Verified at source in this pass:

- `games/oot/soh/GameExports_SingleExe.cpp:1139-1190` — freeze + slot publish, no `.sav` write on the crossing.
- `shared_items.c:271-289` — `RSBS_SHARED_ITEM_REDEEMED` set at award time, entry skipped forever afterward.
[...]

Issue #533, opened 2026-07-29
A refused .redsave (CRC-corrupt, truncated, or newer-version) is indistinguishable from no .redsave and is overwritten at the player's next OoT save with an all-zero Tier-3, unrecoverably destroying the MM half that exists only in that file


spencerduncan:
A `.redsave` that `RsbsSave_Load` refuses — CRC mismatch, truncation, bad inner magic, or an unreadable/newer-version header — leaves the running session in a state that is byte-for-byte identical to "this slot has no cross-game state at all". The only signal is an `fprintf` to stderr that a GUI-launched player never sees. The next OoT save then rewrites the slot unconditionally with a blank Tier-1 and an all-zero Tier-3, destroying the refused file. Since #527 gave MM its redship-native save path, Tier-3 is MM's *only* persistence, so this now converts a single flipped byte into total, unrecoverable loss of MM progress.

## Mechanism

At `f4ad479b`:

1. `games/oot/soh/SaveManager.cpp:170-201` — the `OnLoadFile` hook calls `RsbsSave_SetActiveSlot(fileNum)`, then `Context_InvalidateSessionOnSlotLoad()` (`:197`), which memsets both shadows and `gComboCtx` (`src/common/context.cpp:377-383`). It then gates the reload on `RsbsSave_HasSave(fileNum)` and **discards `RsbsSave_Load`'s `bool`** at `:199`.
2. `RsbsSave_HasSave` (`src/common/save.cpp:344-354`) is header-only by design — it returns `DeserializeHeader(...)`. So a header-level refusal (including a forward version bump) makes `HasSave` false and the load is never even attempted; a body-level refusal makes `Load` return false and nobody looks.
3. Every refusal path in `SaveManager::Load` returns `false` after an `fprintf(stderr, ...)` and nothing else: Tier-1/2/3 truncation (`save.cpp:275-288`), CRC mismatch (`:300-303`), inner `ComboContext` magic (`:307-310`). The commit block at `:312+` is simply skipped.
4. Net effect: post-`OnLoadFile`, a refused slot and an absent slot are indistinguishable — cleared shadows, cleared `gComboCtx`, active slot set.
[...]

Issue #525, opened 2026-07-28
Shared cross-game resources: rupees + hearts (required), hookshot + ammo upgrades (optional) — and the pool shrink they justify


spencerduncan:
Operator directive (during #510 / PR #524 review): *"some items need to be shared which will also cause the item pool to shrink! hookshot; ammo upgrades; RUPEES AND HEARTS!!! Rupees and hearts are the only required share though, the rest can be optional / added later."* Reference: how [OoTMM](https://github.com/OoTMM/OoTMM) does it.

**Required:** shared rupees, shared hearts. **Optional / later:** hookshot, ammo capacity upgrades.

Split out of PR #524 deliberately: that PR is presentation + the reverse pool and carries no save-state risk. This one carves `gComboCtx`, writes both games' live `gSaveContext`, hooks both sides of the switch boundary, and touches MM's cycle reset and the #487 owl-save flash-readback class — the highest-risk area in this repo by its own history. Bundling them would make a red CI un-bisectable.

**The pool shrink belongs HERE, not in #524.** It is *justified by* the sharing: removing the wallet/heart rows before shared resources exist is a strict regression — those items would just vanish from the cross-game pool with nothing replacing them.

## How OoTMM does it (research summary)

Two independent halves. **Pool side** (generation): ~69 `items.shared` boolean settings in `packages/core/src/settings/data.ts`, all defaulting to `false`; `setupSharedItems()` in `packages/logic/src/world/transform.ts` replaces the two games' separate copies with one shared item and removes the duplicate — *this is where the pool shrinks*. **Runtime side**: shared state is stored once and both games read it.

Notably OoTMM has **no "it will be awarded there" concept at all** — one ROM, both save files resident in RAM, one unified GetItem table compiled into both builds, so a cross-game item is simply given where you find it. Our deferred-give model is a consequence of being two separately-booted games in one process, not a design choice to preserve.

### Storage — kind-tagged slots, not bare scalars

Carve from the FRONT of `ComboContext.reserved[]` (`src/common/context.h:441`), directly after `mmProfileDigest`, at literal `.redsave` offset **792**:

```c
#define RSBS_SHARED_RESOURCE_CAP 8u
typedef struct {
    uint8_t  kind;   // RSBS_SHARED_RES_*; 0 == empty slot
    uint8_t  flags;  // RSBS_SHARED_RES_F_MONOTONIC
    uint16_t value;
} ComboSharedResource;                                          // 4 B
[...]

Issue #467, opened 2026-07-21
Audit: every MM entry path that re-evaluates IS_RANDO COND_HOOKs, classified (follow-up to #439/#447)


spencerduncan:
Follow-up audit to #439/#447. #447 fixed the cross-game arrival path and flagged that other MM entry paths "likely share the shape". This issue records the full enumeration so the analysis survives regardless of what gets fixed.

## The bug class (recap)

MM's rando behavior hooks are `COND_HOOK`s gated on `IS_RANDO`. They are re-evaluated **only** by a `GameInteractor_ExecuteOnSaveLoad` dispatch (`Rando.cpp` `OnSaveLoadHandler` -> `Rando::MiscBehavior::OnFileLoad` / `ActorBehavior::OnFileLoad` / `ClockShuffle::OnFileLoad` + `ShipInit::Init("IS_RANDO")`). A dispatch against a **vanilla** save *unregisters* every rando hook. So the question for each entry path is: **what save is live at the moment of dispatch, and is there a later dispatch against the real save?**

No save-file forensics can surface a failure here — the save data is fine, only the hooks are gone.

## Every `OnSaveLoad` / `OnSaveInit` dispatch site in the single-exe build

| # | dispatch site | hook | save live at dispatch |
|---|---|---|---|
| 1 | `games/mm/src/overlays/gamestates/ovl_opening/z_opening.c:32` (`TitleSetup_SetupTitleScreen`) | OnSaveLoad | **vanilla bootstrap** — `MM_Sram_InitNewSave()` at `z_opening.c:23` |
| 2 | `games/mm/src/overlays/gamestates/ovl_file_choose/z_file_choose_NES.c:2250` (`FileSelect_LoadGame`) | OnSaveLoad | the player's real save (`MM_Sram_OpenSave` at `:2201`) |
| 3 | `games/mm/src/code/z_play.c:2416` (`MM_Play_ConsumeStartupEntrance` tail) | OnSaveLoad | the arrival/restored save — **added by #447** |
| 4 | `games/mm/src/code/z_sram_NES.c:1944` (`MM_Sram_InitSave`) | OnSaveInit | the brand-new file being created |
| 5 | `games/mm/src/overlays/gamestates/ovl_select/z_select.c:42` (`MapSelect_LoadGame`) | OnSaveLoad | debug save, gated `fileNum == 0xFF` |
[...]

PR #543, opened 2026-07-29
Stop MM's save-and-quit exits from authoring a bootstrap over the save (#532)

### LINKED ISSUE #532: An owl save in a cross-game MM session routes MM through TitleSetup, which authors a vanilla bootstrap over the live session and arms overwrite of the save just written
Saving at an owl statue during a cross-game MM session sends MM through `MM_TitleSetup_Init`, which authors a fresh vanilla save over the live session. There is no cross-game gate on that branch and no MM-side session invalidation, so `gComboCtx` and OoT's frozen blob stay resident while MM silently becomes an unrelated vanilla world. The next hop back to OoT freezes that bootstrap into the MM shadow, and OoT's next save writes it into Tier-3 — destroying the good MM half the owl save had just persisted moments earlier.

## Mechanism

1. `MSGMODE_OWL_SAVE_1` captures the MM half correctly — `MM_Combo_CaptureSaveToUnifiedSlot` at `games/mm/src/code/z_message.c:6254` — then sets `gSaveContext.gameMode = GAMEMODE_OWL_SAVE` at `games/mm/src/code/z_message.c:6278`.
2. On transition completion, `games/mm/src/code/z_play.c:815-817` does `STOP_GAMESTATE` + `SET_NEXT_GAMESTATE(&this->state, MM_TitleSetup_Init, sizeof(TitleSetupState))`. This branch is ungated — it does not consult `gComboCtx` or any cross-game session flag.
3. `TitleSetup_SetupTitleScreen` (`games/mm/src/code/z_opening.c:16-38`) calls `MM_Sram_InitNewSave()`: a fresh vanilla file becomes the live `gSaveContext`. It then dispatches `OnSaveLoad` against that bootstrap, which unregisters every `IS_RANDO` `COND_HOOK` — same COND_HOOK re-evaluation class tracked in #467.
4. The restore point cannot fire on the way back in. `MM_Play_ConsumeStartupEntrance` early-returns because the arrival already ran `Combo_ClearStartupEntrance` (`games/mm/src/code/z_play.c:2291` / `:2407`) and the frozen MM blob was retired at consume time.
5. There is no MM twin of `Context_InvalidateSessionOnReturnToTitle` — only `games/oot/src/code/title_setup.c` calls it. So the combo context survives a transition that has invalidated the world it describes.
6. Pressing F10 back to OoT (`rsbs/main.cpp:679-681`) runs `Switch_PrepareHotSwap`, freezing the vanilla bootstrap as MM's cross-game state. OoT's next `OnSaveFile` (`soh/SaveManager.cpp:151-168`) serializes that MM shadow into Tier-3, replacing the good MM half.

## Evidence

Verified at source at f4ad479b:

- `games/mm/src/code/z_play.c:808-829` — the `GAMEMODE_OWL_SAVE` branch, confirmed present and with no cross-game condition.
- `games/mm/src/code/z_message.c:6231-6285` — capture at `:6254`, `gameMode` assignment at `:6278`.
- `games/mm/src/code/z_opening.c:16-38` — `MM_Sram_InitNewSave` followed by `OnSaveLoad` dispatch against the bootstrap file.
- `games/mm/src/code/z_play.c:2286-2311` and `:2407` — the startup-entrance presence gate, the `Combo_ClearStartupEntrance` call, and the consume-time retire of the frozen blob.
- `soh/SaveManager.cpp:151-168` — `OnSaveFile` serializes the MM shadow into Tier-3 unconditionally.
- `rsbs/main.cpp:679-681` — F10 back to OoT is live, so the trap is reachable by an ordinary player.
- Absence check: `Context_InvalidateSessionOnReturnToTitle` has exactly one caller, in `games/oot/src/code/title_setup.c`.

## Failure scenario

A player crosses OoT -> Happy Mask Shop -> MM, plays a cycle, and saves at an owl statue. The capture writes a good `.redsave`. MM then restarts through `TitleSetup`: the player is dropped at MM's vanilla title/opening screen on a blank file with all rando hooks unregistered — their session apparently gone. Pressing F10 to return to OoT, the only graceful route out, freezes the blank bootstrap as MM's cross-game state. OoT's next save (autosave fires within three minutes) writes it into Tier-3, permanently destroying the MM save the owl statue had just persisted.

## Severity

P0 — silent, unrecoverable loss of the MM half of a cross-game save, triggered by an ordinary in-game save action with no warning to the player.

Found by the 2026-07-29 multi-agent bugbash (static sweep at f4ad479b); confirmed by adversarial re-derivation from source.

spencerduncan:
Fixes #532.

## The destruction

MM's owl-save and pause save-and-quit exits both do
`SET_NEXT_GAMESTATE(MM_TitleSetup_Init)` from inside MM's own graph loop, with
**no cross-game condition**. `TitleSetup_SetupTitleScreen` calls
`MM_Sram_InitNewSave()`, so a fresh vanilla file becomes the live
`gSaveContext` — while `gComboCtx` and OoT's frozen blob stay resident, because
MM has no twin of `Context_InvalidateSessionOnReturnToTitle`.

The next hop back to OoT freezes that bootstrap into the MM shadow, and OoT's
next save serializes it into Tier-3 — **destroying the MM half the save
persisted seconds earlier.** The player watches a successful save silently
become an unrelated vanilla world.

## The fix

`MM_Combo_OwlSaveExitToOoT()` requests a launcher switch instead of entering
`TitleSetup`. `Combo_CheckHotSwap` returns `Combo_IsGameSwitchRequested()`, so
MM's graph loop breaks and control returns to `rsbs/src/main.cpp`, which takes
its `!isEntranceSwitch` branch: freeze MM's live (correctly saved) state, set
OoT's startup entrance from OoT's own frozen return entrance, switch.

**MM never reaches `TitleSetup`, so the bootstrap is never authored and there is
nothing to launder.**

Gated on `Context_HasFrozenState(GAME_OOT)` — precisely *"there is an OoT
session to return to"*. A standalone MM session has none and keeps vanilla 2ship
behavior; this must not hijack upstream's exit.

Applied to both **save** exits: owl statue (`z_play.c`) and pause save-and-quit
(`z_kaleido_scope_NES.c`).

## Deliberately not changed

The game-over *"don't continue"* prompt at `z_kaleido_scope_NES.c:3987` takes
the same `TitleSetup` escape — but it's a **death** path, not a save path, and
outside this issue's scope. Changing death semantics without analysing them
separately is how you trade one bug for a worse one. It wants its own change,
likely alongside #530.

## Scope limit

This returns the player to OoT **gameplay** at the frozen return entrance (the
Happy Mask Shop they left from), not to OoT's **menu**. The menu half is the
remaining part of "one game one menu". Fixing the destruction first, because
that's the part that loses the player's progress.

## Verification

Local: **63/63 redship, 14/14 rando.** Full ROM-staged Ninja/Release build. No
format change.

No new test: the behavior is a gamestate-transition decision inside MM's graph
loop, which the headless harness has no way to drive — asserting on
`Combo_IsGameSwitchRequested()` alone would test the one-line helper rather than
the escape it prevents, i.e. a test that passes whether or not the call site is
wired. Calling that out rather than shipping a vacuous one. The reachable
regression guard is the existing `MMOwlSaveArmState` row plus manual
verification of the owl-save round trip.

### Build Artifacts
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8829534069.zip)
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8829411040.zip)


spencerduncan:
## Adoption pass: rebased, adversarially reviewed, locked

The session that opened this PR went inactive, so this is an adoption pass by another
session. **I pushed to this branch** (`f457a617..2b96114e`) rather than opening a
competing PR — the work is being brought home, not replaced.

**Recommendation: keep this PR open. The fix is required and is NOT superseded.**
It is not merge-ready yet, for a reason that has nothing to do with the change — see
*Verification* at the bottom.

---

### 1. Rebase: merged current `origin/main` (`5b23bf28`)

Clean, no conflicts. That brings in the whole 2026-07-29..31 save-layer arc — the
libultraship archive mutex (#566), consumer guards (#561), and alignment phase 1
(#567/#568/#569/#570: REFUSED state, the commit choke point + generations, profile
freeze).

### 2. Is #532 already fixed by the phase-1 machinery?

This was the main thing worth checking, since the bootstrap-authoring path now
coexists with a write latch, commit generations and freeze refusal. **It is not.**
Verified against `origin/main` rather than assumed:

| Claim | Evidence |
|---|---|
| The mechanism is still live | `games/mm/src/code/z_play.c:821` still takes the **ungated** `SET_NEXT_GAMESTATE(&this->state, MM_TitleSetup_Init, ...)` on `GAMEMODE_OWL_SAVE` |
| The freeze does not screen it | `Switch_PrepareHotSwap` (`src/common/switch.cpp:84`) validates only the game id, the pointer and the size — **nothing about the plausibility or identity of the bytes**. A vanilla bootstrap freezes exactly as happily as a real session |
| The commit choke point does not screen it | `SaveManager::Save` is latch-check + stage + write; #569 made the commit atomic and generation-stamped, it did **not** make it validated |
| The #568 latch is satisfied, not violated | `RsbsSave_IsSlotWritable` is armed by load/create/erase — **the owl capture itself arms the write that then destroys it** |
| #570 is too late | `MM_Rando_PairOnCrossGameArrival` runs at **arrival** (`z_play.c:2350`), so it can only detect the corruption on a later boot — after the good Tier-3 is already gone |

The fix **composes** with the new machinery rather than bypassing or duplicating it:
it performs no save of its own (the owl capture already went through
`RsbsSave_Save` → `StageCommit`/`WriteStagedCommit` on MM's game thread), and it does
not touch the MM shadow (`Switch_PrepareHotSwap` stays the single writer on this path).

### 3. Adversarial review findings (all addressed in `2b96114e`)

**A. The kaleido leg is dormant, not a second live fix.** Nothing in `games/mm` ever
assigns `PAUSE_SAVEPROMPT_STATE_6` — the only occurrences are the `case` label
(`z_kaleido_scope_NES.c:3736`), a read at `:1023`, and the enum definition. MM's
*reachable* pause save is `PAUSE_SAVEPROMPT_STATE_1` (where the pre-existing
`MM_Combo_CaptureSaveToUnifiedSlot` call lives at `:3619`); it **saves and returns to
gameplay** via `STATE_3` → `PAUSE_STATE_UNPAUSE_CLOSE`. It does not quit. So the PR
body's *"applied to both save exits"* overstates. I kept the guard — the state could
be wired up later and silently re-arm the mechanism — and rewrote the comment to say
so honestly instead of implying a second live exit.

**B. The new helper orphaned a doc block.** It had been inserted between
`MM_Combo_CaptureSaveToUnifiedSlot`'s doxygen comment and the function itself, so
those docs landed on the wrong symbol. Moved below.

**C. No re-trigger loop — but that is load-bearing, not lucky.** Skipping the
transition leaves `gameMode == GAMEMODE_OWL_SAVE` in the frozen blob. What stops the
next arrival into MM from immediately re-running the same exit is
`MM_Play_ConsumeStartupEntrance` (`z_play.c:2363`), which resets it to
`GAMEMODE_NORMAL`. Deleting that one line turns this fix into a loop. Now documented
at the helper.

**D. The gate's meaning depends on #364.** `Context_HasFrozenState(GAME_OOT)` means
*"there is an OoT session to go back to"* only because `Combo_ConsumeFrozenState`
(`src/common/switch.cpp:123`) retires the blob as it applies it — otherwise a frozen
state here could be a stale souvenir of a trip the player already came back from.
Documented; that invariant is itself locked by `src/common/tests/test_hotswap_freeze.c`.

**E. The game-over "don't continue" path is a KNOWN LIVE HOLE.** It carries the
*identical* mechanism (`z_kaleido_scope_NES.c`, `PAUSE_PROMPT_NO` → `MM_TitleSetup_Init`).
Excluding it from this PR is the right call — it is a **death** path, and combo death
semantics deserve their own analysis rather than a transcription of the owl-save
answer. But it is not a cleared path, so I marked it `RSBS-TITLESETUP-EXEMPT(#532)`
with the reasoning written out, which forces the new invariant to **name it** rather
than let it hide among the legitimately-exempt front-end transitions. Worth its own
issue.

### 4. Locks (the PR originally shipped none)

The original objection was correct: asserting on `Combo_IsGameSwitchRequested()` alone
tests the one-line helper, not the escape it prevents — it passes whether or not the
call site is wired. So the lock is **split**, and neither half is vacuous alone.

**Lock 1 — DECISION.** `games/mm/2s2h/mm_unified_save_test.cpp` section 6, folded into
the **existing** `MMUnifiedSaveCapture` row (no new CMake row). Deliberately in the
file that already writes a good MM half, so the same test proves the save is written
and then locks the exit that used to eat it:

- **6a standalone MM** → returns `0` and requests no switch.
  *Counterfactual:* an implementation that drops the `Context_HasFrozenState` gate
  ("always switch") hijacks upstream 2ship's own save-and-quit and strands a
  standalone player in a game they never asked for — this row goes red.
- **6b cross-game MM** → returns `1` **and** `Combo_IsGameSwitchRequested() == true`.
  *Counterfactual:* returning `1` without requesting the switch is **inert** — MM
  keeps running frames and reaches TitleSetup anyway. That variant passes a naive
  "did it return 1" check and fails this one.
- **6c capture refused by the #533 latch** → the exit **still** returns `1` and still
  requests the switch.
  *Counterfactual:* an implementation that gates the exit on "did the save land"
  sends the player into TitleSetup precisely when the durable half is already lost —
  the worst case — and fails here.

**Lock 2 — WIRING.** `tools/tests/test_repo_invariants.py::test_titlesetup_transitions_are_guarded_or_marked_exempt`
(runs in `python-tests.yml`, which sweeps `tools/tests/` unfiltered). Every
`SET_NEXT_GAMESTATE(..., MM_TitleSetup_Init, ...)` under `games/mm/src` must be
guarded, marked `RSBS-TITLESETUP-EXEMPT` with a reason, or in MM's own front end
(`z_title.c` / `z_file_choose_NES.c`, where there is no live session to destroy).

*Why it has to be source text:* nothing in the C build can fail on a **new** ungated
transition. It compiles, links, and passes every ctest row. The fix is a call-site
decision, so only a source-text assertion can see its removal.

A companion test (`test_titlesetup_invariant_sees_the_known_transitions`) pins that
the scanner still matches, so a rename or macro change cannot make Lock 2 pass while
enforcing nothing.

### 5. Verification — please read before merging

**Ran and green:**
- `tools/tests/test_repo_invariants.py` — **12/12**, including both new invariants.
- **Counterfactual re-run this pass:** replacing `if (!MM_Combo_OwlSaveExitToOoT()) {`
  with `if (1) {` in `z_play.c` fails **both** new invariants *independently* — the
  first naming the file and line, the second dropping the guarded count 2 → 1 (so
  they do not fail in lockstep off a single root cause). Restored via the editor
  (not a shell round-trip, which previously corrupted this file's UTF-8) and
  re-confirmed 12/12 green with an identical diffstat.
- Every API the new test uses re-checked against **real** declarations on current
  main: `Context_FreezeState(GameId, uint16_t, const void*, size_t)` (`context.h:53`),
  `RsbsSave_ResetSlotSessionState` / `RsbsSave_SetActiveSlot` (`save.h:574,581`),
  `Combo_*GameSwitch*` inside `entrance.h`'s `extern "C"` block (`:109-111`),
  `OOT_ENTR_MARKET_FROM_MASK_SHOP` (`entrance.h:28`), `OOT_SAVE_CONTEXT_SIZE`.
  6c's premise re-derived from the implementation: `ResetSlotSessionState` disarms
  all slots (`save.cpp:870`), `IsSlotWritable` is `mSlotArmed[slot]` (`:787`), and
  `Save` returns false on `CheckWriteAllowed` (`:273`) — so the capture really is
  latch-refused there. Section 6 also leaves no global state behind: the pre-existing
  cleanup resets the active slot to `-1` and the latch state.

**NOT run: the full ROM-staged build and both ctest tiers.** The machine has
**4.4 MB free on C:** — a build measures ~13.3 GB, so it is physically impossible
right now. **The C++ additions in `mm_unified_save_test.cpp` are UNCOMPILED.**

Roughly 150 GB of stale `build-cmake` output from completed 7/30–7/31 workflow runs
is sitting in `.claude/worktrees`, and agents are blocked from reclaiming it. This is
an environment blocker affecting every lane, not something specific to this change.

Under the operator's standing local-verification gate, **this branch is not
merge-ready**: someone with disk headroom must run the ROM-staged Ninja/Release build
and both tiers (`ctest -L "^redship$"`, then `-L "^rando$"`) first. Pushed so CI can
compile it in the meantime. **I did not merge, and I am not recommending a merge yet.**

### 6. Residuals (not fixed here)

- **One-game semantics.** An owl save means *"save and quit"*, and this delivers
  *"save and teleport into OoT gameplay at Hyrule Market"*. Per the one-game ruling
  the combo is ONE game, so quitting should land at the combo's front end. The
  ordering is right — losing the save is strictly worse than landing in the wrong
  place — but the menu half needs `Context_InvalidateSessionOnReturnToTitle`
  semantics plus a launcher signal that suppresses the startup-entrance branch at
  `rsbs/src/main.cpp:709`. Wants its own issue.
- **Pre-existing, untouched:** `MM_Combo_CaptureSaveToUnifiedSlot` runs
  `MM_HarvestSharedResources()` *before* `RsbsSave_Save`, so a latch-refused capture
  still advances the in-memory shared pool for a commit that never lands — relevant
  to the harvest/apply gate-symmetry class.
- **Incidental closure:** because MM never reaches TitleSetup, the `OnSaveLoad`
  dispatch against the bootstrap (the `IS_RANDO` `COND_HOOK` unregistration of #532
  step 3 / #467) never happens either.

spencerduncan:
## Second-signature verification pass (no code change)

A follow-up adoption lane picked this PR up after the comment above. The branch was
already rebased onto current `origin/main` (`5b23bf28`) and pushed at `2b96114e`, so
**there was nothing left to reconcile and I pushed no new commits.** What I did
instead was re-derive the load-bearing premises independently rather than inherit
them, and attack the one gap the previous pass flagged: the C++ additions are
**uncompiled**.

**Recommendation unchanged: keep this PR open, do not merge yet.** The change is
correct as far as static verification can reach; the local-build gate is still
blocked by the environment, not by the diff.

---

### 1. Premises re-derived against the tree (not taken on trust)

Each of these was checked from the source on this branch, and each one, if false,
would break the fix:

| Premise the fix rests on | Independently confirmed |
|---|---|
| The owl save actually persists before the exit | `games/mm/src/code/z_message.c:6254` calls `MM_Combo_CaptureSaveToUnifiedSlot()` **outside** the `fileNum != 0xFF` gate — so in a cross-game session (fileNum pinned to `0xFF`) that call is the *only* thing that persists anything. The save this PR protects is real |
| Requesting a switch actually stops MM | `Combo_CheckHotSwap()` is polled at the **top** of `MM_Graph_ThreadEntry`'s loop *before* `MM_RunFrame` (`games/mm/src/code/graph.c:523`), so the loop breaks at the very next frame boundary |
| …and the poll reports this request | `Combo_CheckHotSwap` (`games/oot/soh/GameExports_SingleExe.cpp:1720`) falls through to `Combo_IsGameSwitchRequested()`, which is the flag `Combo_RequestGameSwitch` sets (`src/common/entrance.cpp:342`). Nothing on that path clears it |
| The launcher then lands the player in OoT gameplay | `rsbs/src/main.cpp:602` derives `nextGame = Game_GetOther(active)` with `isEntranceSwitch = false`, and `:709` takes `!isEntranceSwitch && Context_HasFrozenState(nextGame)` → `Entrance_SetStartupEntrance(Context_GetFrozenReturnEntrance(...))`. **The launcher gates on the same predicate the fix does**, so the fix cannot request a switch that the launcher then fails to route |

**Dropping `STOP_GAMESTATE` is not a divergence.** I checked this specifically, since
the guarded branch skips it. `STOP_GAMESTATE` does not short-circuit the current
frame in the vanilla path either — the remainder of `Play_Update` runs eitherway, and
the gamestate is torn down at the frame boundary. Here the graph loop breaks at that
same boundary, and `PlayState` is discarded rather than preserved
(`MM_Graph_ResetRunFrameContext` — cross-game continuity rides the frozen
`SaveContext`, not the gamestate). So the leftover `transitionMode` is unreachable.

### 2. Compile-risk audit of the uncompiled C++

Since the ctest rows could not be built, I audited the new TU statically for the
failure mode that would actually bite — the `#ifndef __cplusplus` trap, where a
declaration greps as in-scope but is invisible to a C++ TU. **All six symbols the new
code touches resolve inside live `extern "C"` blocks:**

- `Context_HasFrozenState` (`context.h:69`) and `Context_ClearAllFrozenStates`
  (`:84`) — inside the block opened at `:29`, closed at `:1123`.
- `Combo_RequestGameSwitch` (`entrance.h:109`), `Combo_IsGameSwitchRequested`
  (`:110`), `Combo_ClearGameSwitchRequest` (`:111`) — inside `:78`–`:114`.
- `OOT_ENTR_MARKET_FROM_MASK_SHOP` (`entrance.h:28`) is a `#define`, so its position
  outside the block is irrelevant.

`GameExports_SingleExe.cpp` already includes both `context.h` and `entrance.h`, and
already calls `Context_HasFrozenState(GAME_MM)` elsewhere in the same TU — so the new
helper adds no new include or linkage requirement. `mm_unified_save_test.cpp` already
includes `<vector>` (`:82`), which its new `std::vector<uint8_t> ootSession` needs.

This is not a substitute for compiling it. It does mean there is no *known* break.

### 3. Locks re-run independently, including the counterfactual

- `tools/tests/test_repo_invariants.py` — **12/12 green** on this branch.
- **Counterfactual re-run from scratch:** I deleted the
  `if (!MM_Combo_OwlSaveExitToOoT())` wrapper in `z_play.c` outright (a true revert to
  main's shape, not a `if (1)` stub). Both new invariants failed, and failed
  *independently* — `test_titlesetup_transitions_are_guarded_or_marked_exempt` naming
  the now-unguarded `z_play.c` transition, and
  `test_titlesetup_invariant_sees_the_known_transitions` separately dropping the
  guarded count `2 → 1` (leaving only the kaleido site). Two distinct assertions, so
  they do not fail in lockstep off one root cause. Restored via `git checkout --`
  (byte-exact, avoiding the shell round-trip
  that previously corrupted this file's UTF-8); tree clean, 12/12 green again.

So Lock 2 is confirmed non-vacuous by a second party. Lock 1 (the C++ decision rows)
remains **unexecuted** — see below.

### 4. Why the build still did not run

Same blocker, unchanged and not specific to this PR: **C: has 0.44 GB free.** A
ROM-staged build needs ~13 GB.

I surveyed the cause — **~98 GB of `build-cmake` output across 18 agent worktrees**,
almost all of it from workflow runs that finished on 7/30–7/31 (twelve directories at
~6.6 GB each). Every one is still git-registered, so registration does not
discriminate live from dead; last-write time is the only signal, and by it exactly two
are plausibly in use right now (`wf_4221f5cf-c9d-2` and `-3`, written within the last
half hour).

**I did not reclaim any of it.** These are other sessions' working directories, and
with two lanes apparently mid-build, deleting the wrong tree breaks a peer's run. That
call belongs to the operator, not to a lane agent. Reclaiming the twelve 7/30–7/31
`build-cmake` directories — *excluding* `wf_4221f5cf-c9d-*` and any currently-running
lane — would free ~80 GB and unblock local verification for every lane at once.

**Status: `Lock 1` unexecuted, both ctest tiers unexecuted, C++ additions uncompiled.**
Under the standing local-verification merge gate this branch is **not merge-ready**,
for environmental reasons only. CI compiles the branch in the meantime. I did not
merge and am not recommending a merge.

---

### Update: CI finished green after this comment was written

`build-linux` **and** `build-windows` both went `SUCCESS` on `2b96114e`, along with
`clang-format`, `clang-tidy`, `odr-declaration-gate` and the python `test` row.

**That retires the "the C++ additions are UNCOMPILED" caveat** — the new
`MM_Combo_OwlSaveExitToOoT` helper, both call sites, and the section-6 additions to
`mm_unified_save_test.cpp` compile and link on two toolchains. The static audit in
§2 above predicted exactly that; CI confirmed it.

**What is still outstanding is narrower than before, and it is one thing:** Lock 1's
rows have been *compiled* but never *executed*. No CI check runs the ctest tiers —
the `test` row is `python-tests.yml`, i.e. Lock 2 only. So the runtime assertions in
section 6 (6a standalone, 6b cross-game, 6c latch-refused) have not been observed
passing anywhere, and neither has the rest of the `redship` / `rando` tiers against
this branch.

Revised status: **the only remaining gate is `ctest -L "^redship$"` and `-L "^rando$"`
on a machine with disk headroom.** Everything else on this PR is now verified.

spencerduncan:
## Third-signature adversarial review (no code change, no new commits)

Adoption lane again. Branch head is unchanged at `2b96114e`, `origin/main` has not
moved since (`5b23bf28`, and #581 is still open), so there was nothing to re-merge.
I pushed nothing. This pass exists to attack the parts the two comments above
*asserted* rather than *exercised*.

**Recommendation unchanged: keep open, do not merge.** The fix is not superseded by
the phase-1 machinery, and I re-derived that independently below.

### 1. A counterfactual the earlier passes did not run

Both earlier passes falsified Lock 2 the same way: delete the existing
`if (!MM_Combo_OwlSaveExitToOoT())` wrapper in `z_play.c` and watch the invariant go
red. That proves it catches a *regression at a known line*. It does **not** prove the
property the docstring actually claims — that a **future edit adding a new ungated
transition** is caught.

So I ran that one. I added a brand-new file
`games/mm/src/code/z_rsbs_counterfactual.c` containing nothing but an unguarded
`SET_NEXT_GAMESTATE(&play->state, MM_TitleSetup_Init, sizeof(TitleSetupState))` —
touching no existing file, so there is no encoding round-trip risk and no chance of
disturbing the guard under test:

```
E  AssertionError: MM_TitleSetup_Init transitions with neither a MM_Combo_OwlSaveExitToOoT
   guard nor a RSBS-TITLESETUP-EXEMPT marker (#532 -- these author a vanilla bootstrap
   over a live cross-game session): ['games/mm/src/code/z_rsbs_counterfactual.c:3']
1 failed, 11 passed in 0.65s
```

Removed the file; `12 passed`, `git status --porcelain` empty. **Lock 2 enforces the
forward-looking property, not just the historical one.** That is the half that matters,
because the C build cannot fail on a new ungated transition.

### 2. A composition check nobody had made — and it is the one that could have bitten

The fix deliberately skips `STOP_GAMESTATE`, which means the launcher's hot-swap freeze
captures MM with `gSaveContext.gameMode == GAMEMODE_OWL_SAVE` still resident. The
earlier comments treat that as inert. It is not inert: `gameMode` is the input to the
#525 live-file gate.

`src/common/shared_resources.c:186`:

```c
return gameMode == RSBS_GAMEMODE_NORMAL || (game == GAME_MM && gameMode == RSBS_GAMEMODE_OWL_SAVE);
```

`OWL_SAVE` is an **accepted** live-file mode for MM (and only for MM —
`test_shared_resources.c:580` pins that OoT must reject the same numeric value), and
`GameExports_SingleExe.cpp:2418` static-asserts the enum values agree. So the
switch-time shared-resource harvest still runs on this departure.

**Had `OWL_SAVE` not been in that predicate, this PR would silently have converted
every owl-save departure into a harvest-less one** and walked the shared pool out of
sync with the MM half the save just persisted — a data bug with no failing test
anywhere. It composes correctly, but by agreement with a predicate written for another
issue, not by construction. Worth stating out loud so a future edit to
`Combo_SaveIsLiveFile` knows this path depends on it.

### 3. One mechanism correction to the comment above

> `PlayState` is discarded rather than preserved (`MM_Graph_ResetRunFrameContext`)

The conclusion is right; the mechanism is not. `MM_Graph_ResetRunFrameContext`
(`games/mm/src/code/graph.c:542`) only NULLs `runFrameContext` and `MM_gPlayState` — it
calls neither `MM_GameState_Destroy` nor `Play_Destroy`. On this path, as on **every**
F10 departure, the retired session's allocations are abandoned in the system arena and
`MM_Game_Resume` re-arms the arena on the next entry.

That is the stronger version of the argument, not a weaker one: the fix does not
invent a teardown path. It departs exactly the way an F10 press already departs —
mid-frame, with a live `PlayState`, without `Play_Destroy`. The only new thing is
*when* (mid-transition, `TRANS_MODE_INSTANCE_RUNNING`), and since nothing on the
departure path inspects `transitionMode`, the leftover value is unreachable.

### 4. Premises re-derived a third time, independently

- `games/mm/src/code/z_message.c:6254` — `MM_Combo_CaptureSaveToUnifiedSlot()` sits
  **outside** the `fileNum != 0xFF` gate; in a cross-game session it is the only thing
  that persists anything. Confirmed by reading the site, not the comment.
- Same state machine sets `gSaveContext.gameMode = GAMEMODE_OWL_SAVE` in
  `MSGMODE_OWL_SAVE_1` for **both** standalone and cross-game sessions, so the guarded
  branch in `Play_UpdateTransition` is reached identically in both and
  `Context_HasFrozenState(GAME_OOT)` is the *sole* discriminator. That is the right
  discriminator: under #364's single-use blob contract, `Combo_ConsumeFrozenState`
  retires OoT's blob on arrival back in OoT, so "OoT is frozen" cannot be a stale
  souvenir — it means "the player is inside MM having come from OoT", which is exactly
  the condition.
- `graph.c:521-527` — `Combo_CheckHotSwap()` is polled at the top of
  `MM_Graph_ThreadEntry`'s loop and `break`s. And note the fix does not *race* it:
  because `SET_NEXT_GAMESTATE` is never called, `TitleSetup` is not scheduled at all,
  so even an arbitrarily delayed break could not reach it.
- `GameExports_SingleExe.cpp:1720-1745` (OoT side) — `Combo_CheckHotSwap` falls
  through to `Combo_IsGameSwitchRequested()`, the flag `Combo_RequestGameSwitch` sets;
  nothing on this path clears it.
- `rsbs/src/main.cpp:600-602, 709` — the request lands in the `!isEntranceSwitch`
  branch and gates on `Context_HasFrozenState(nextGame)`, **the same predicate the fix
  uses**, so the fix cannot request a switch the launcher fails to route.

### 5. Dissent I am recording rather than fixing

Under the one-game ruling, an owl save is a **half-file commit**: it captures MM's
half, while the OoT half of the same paired file stays at whatever OoT last durably
wrote — and this fix then hands the player back the *live, uncommitted* OoT session at
Hyrule Market. "Save and quit" in one game therefore neither saves the combo nor
quits it.

None of that is introduced here — the capture is already on `main`, and this PR
strictly improves the outcome (losing the MM half is worse than landing in the wrong
place). But it is a combo-level semantics gap, distinct from the presentation residual
already disclosed in the PR body, and it wants its own issue rather than being folded
into #532.

### 6. Verification status

- `python -m pytest tools/tests/test_repo_invariants.py -q` → **12/12** on `2b96114e`,
  and 12/12 again after the §1 counterfactual was reverted. Worktree clean.
- CI on `2b96114e`: **all green** — `build-linux`, `build-windows`, `clang-format`,
  `clang-tidy`, `odr-declaration-gate`, `link-check`, `generate-rsbs-otr`,
  `netplay-relay`, `test`.
- **Both ctest tiers: still unexecuted.** Lock 1's rows (6a/6b/6c) compile on two
  toolchains but have never been observed passing. The blocker has *worsened* since the
  comment above: `C:` is now at **0.05 GB free**, down from 0.44 GB. A ROM-staged build
  needs ~13 GB. This is environmental and unrelated to the diff; the reclaim
  recommendation in the previous comment stands and is now urgent.

Not merged, nothing closed, no commits pushed.

spencerduncan:
## Fourth adoption pass: the local gate is CLOSED — both tiers run, and Lock 1 is falsified twice

Adoption lane. Branch head unchanged at `2b96114e`; `origin/main` is still `5b23bf28`
(#581 still open), and the branch already contains it, so **there was nothing to
re-merge and I pushed no commits.** No code changed on this pass.

The three comments above converge on one outstanding item: *"the only remaining gate is
`ctest -L "^redship$"` and `-L "^rando$"` on a machine with disk headroom."* **`C:` now
has 192 GB free**, so I ran it.

**Recommendation: keep open. The last blocking gate is now satisfied.** Not merging —
that is the shepherd's call, not a lane agent's.

---

### 1. Full Ninja/Release build + both tiers — GREEN

Fresh worktree off `2b96114e`, submodules initialized, `-j4` at low priority per the
standing resource cap.

```
cmake -S . -B build-cmake -G Ninja -DCMAKE_BUILD_TYPE:STRING=Release   -> configure ok
cmake --build build-cmake -- -j4                                       -> 3016/3016, EXIT=0

ctest -L "^redship$"   ->  73/73 passed  (ZipContention skipped, as always)
ctest -L "^rando$"     ->  15/15 passed
```

`MMUnifiedSaveCapture` (row 66) — **Lock 1 — has now been executed for the first time.**
It passes.

**One staging gotcha worth recording**, because it will bite the next lane that verifies
a branch in a fresh worktree: the first `rando` run came back **14/15** with
`RandoGenFullInit` red, and it is *not* a regression —

```
[TEST] FAIL: no soh.o2r resolvable (tried './soh.o2r'). This row needs a mounted
archive: with none, libultraship pauses the resource thread pool and OTRAudio_Init's
synchronous load never returns.
```

That row (added by #562) needs the **port** archives in the build root. They are
ROM-free — `cmake --build build-cmake --target GenerateSohOtr Generate2ShipOtr` produced
`soh.o2r` + `2ship.o2r` and the tier went 15/15. No ROMs and no `ExtractAssets` needed
for either tier; that is only the `integration` label.

### 2. Lock 1 falsified — twice, with distinct assertions

Passing is not the interesting part. Nobody had ever shown these rows can go **red**, so
"6a/6b/6c" was still an untested claim about a test. I ran both counterfactuals the PR
body names, each as a one-TU edit to `MM_Combo_OwlSaveExitToOoT` plus an incremental
relink.

**CF-1 — drop the `Context_HasFrozenState(GAME_OOT)` gate** ("always switch", the variant
that would hijack upstream 2ship's own save-and-quit):

```
[TEST] FAIL: a standalone MM session must keep the vanilla TitleSetup exit (#532 gate)
             (games/mm/2s2h/mm_unified_save_test.cpp:207)
```

**CF-2 — return 1 without calling `Combo_RequestGameSwitch()`** (the *inert* variant: it
refuses TitleSetup on paper, but MM keeps running frames and reaches TitleSetup anyway —
exactly the implementation a naive "did it return 1" check would wave through):

```
[TEST] FAIL: refusing TitleSetup is not enough: the exit must request the switch that
             breaks MM's graph loop, or MM keeps running and reaches TitleSetup anyway
             (games/mm/2s2h/mm_unified_save_test.cpp:226)
```

Two different rows, two different assertions, neither failing off the other's root cause.
Restored with `git checkout --`, rebuilt, re-ran **both** tiers: 73/73 and 15/15, worktree
`git status --porcelain` empty (including the regenerated `build.c` / `properties.h`
stamps, which I reverted rather than carried).

So Lock 1 is non-vacuous by execution, not by argument — and together with the third
comment's new-file counterfactual against Lock 2, both halves of the split lock are now
demonstrated to fail on the implementations they claim to exclude.

### 3. Premises I re-derived rather than inherited

- **`PAUSE_SAVEPROMPT_STATE_6` is genuinely dead.** The first comment established there is
  no enum-name assignment; I also checked the two ways that claim could still be wrong —
  a **numeric** assignment (`savePromptState = 6`) and an **increment**
  (`savePromptState++`). Neither exists: every write to `savePromptState` in `games/mm/src`
  is a named-constant assignment, and none of them is `STATE_6`. The kaleido leg really is
  dormant, and the PR's decision to guard it anyway costs nothing.
- **The `MM_Play_Init` exemption is sound by ordering, not by assertion.**
  `MM_Play_ConsumeStartupEntrance()` is called at `games/mm/src/code/z_play.c:2518`; the
  `save.entrance == -1` bail is at `:2536`. A cross-game arrival has assigned a real
  entrance eighteen lines before the check can see it.
- **Nothing clears the switch request between the call and the poll.**
  `Combo_CheckHotSwap` (`games/oot/soh/GameExports_SingleExe.cpp:1720`) returns
  `Combo_IsGameSwitchRequested()` on *every* path including both early returns, and
  `sGameSwitchRequested` is cleared only by `Combo_ClearGameSwitchRequest`
  (`src/common/entrance.cpp:350`), which the launcher calls after it has committed to the
  switch.

### 4. One review finding the earlier passes did not name

**The launcher can REFUSE a switch, and on this path a refusal has no exit.**

`rsbs/src/main.cpp` has four refusal branches that all do
`Combo_ClearGameSwitchRequest(); ... continue;` — missing game archive (`:613`), missing
port archive (`:626`), OoT assets not initialized at boot (`:641`), and a failed hot-swap
freeze (`:667`). `continue` re-enters `while (keepRunning)` and calls `ops->run()` on the
**still-active** game; `MM_Game_Suspend` never ran, so `runFrameContext` is intact and
MM resumes the same frame loop.

At that point `transitionMode` is still `TRANS_MODE_INSTANCE_RUNNING` and
`gSaveContext.gameMode` is still `GAMEMODE_OWL_SAVE` — because the guarded branch
deliberately skips `STOP_GAMESTATE`, and nothing else in that branch advances the
transition state machine. So `Play_UpdateTransition` re-enters the same case, calls
`MM_Combo_OwlSaveExitToOoT()` again, requests again, breaks again, is refused again:
**a request/refuse livelock, where vanilla had a TitleSetup fallback.**

Note this is *specific to this call site*. An F10 refusal is harmless precisely because
the player's keypress is not regenerated every frame; here the request is a property of
parked gamestate, so it is.

**I am recording it, not blocking on it,** because in the #532 scenario every one of the
four refusals is effectively unreachable: you can only *be* in a cross-game MM session if
OoT booted and its archives mounted (killing `:613`/`:641`), MM is currently running from
`2ship.o2r` (killing `:626`), and `Switch_PrepareHotSwap` refuses only on a NULL
SaveContext or zero size (`src/common/switch.cpp:84-92`), which a live MM does not have.
It is a structural note for whoever touches either side next — if a future refusal
condition becomes reachable (a new archive check, a validating freeze), this call site
turns a refusal into a hang rather than a fallback. Cheapest future insurance would be
for the guarded branch to set `transitionMode = TRANS_MODE_OFF` before returning, so a
refused switch drops back into gameplay instead of re-arming.

### 5. Status

| Gate | State |
|---|---|
| Rebase onto current `origin/main` | done (`2b96114e`, still current) |
| CI (both toolchains, format, tidy, ODR gate) | green, per the comment above |
| Lock 2 (source-text wiring invariant) | 12/12, falsified by two independent counterfactuals |
| Lock 1 (runtime decision rows) | **executed and passing; falsified twice** |
| `ctest -L "^redship$"` | **73/73** |
| `ctest -L "^rando$"` | **15/15** |

Nothing merged, nothing closed, no commits pushed.

spencerduncan:
## Adversarial review pass — one lock hole found and closed, both tiers re-verified

Fifth adoption pass on this PR. The previous pass closed the build gate (full local build + both tiers, and the first-ever execution of Lock 1). This pass re-derived the premises independently rather than inheriting them, and **found a real vacuity hole in Lock 2** that four prior passes missed. Pushed one commit: `5b9cb2c4`.

This is another session's branch; the work is being adopted, so the push is deliberate. **Not merged, nothing closed.**

### Rebase status

`origin/main` is still `5b23bf28`. #581 has not landed. `git merge origin/main` → already up to date; branch **MERGEABLE**, no reconciliation needed.

### Not superseded by phase-1 — re-derived at the decisive link

I did not take the prior passes' word for this. The claim that matters is whether the #533/#568 write latch already blocks the *laundering* step, which would make the fix redundant. It does not, and here is the exact reason:

```
src/common/save.cpp:219  SaveManager::CheckWriteAllowed(int slot)
    if (!mSlotArmed[slot]) { ...refuse... }
```

The latch arms on **"this session loaded, created, or erased the slot"** — not per-write. In the #532 scenario the player loaded the slot at boot, so `mSlotArmed` is true for the whole session and OoT's post-return save is **permitted**. The bootstrap reaches Tier-3 exactly as before phase-1.

The freeze leg is equally open — `Switch_PrepareHotSwap` (`src/common/switch.cpp:84-92`) validates only game id, non-NULL pointer, non-zero size. A vanilla bootstrap freezes as happily as a real session. And #570's `MM_Rando_PairOnCrossGameArrival` runs at *arrival*, i.e. after the Tier-3 damage.

**Conclusion: the fix is required and composes rather than duplicates.** It performs no save of its own and does not touch the MM shadow.

Also independently confirmed: the no-re-trigger-loop claim in the helper's docstring is real — `gSaveContext.gameMode = GAMEMODE_NORMAL` at `games/mm/src/code/z_play.c:2363`, inside the cross-game arrival block, after `MM_Play_ConsumeStartupEntrance`.

### NEW FINDING — Lock 2 could be silenced by a comment (fixed in `5b9cb2c4`)

`test_titlesetup_transitions_are_guarded_or_marked_exempt` matched the guard's name as a **raw substring** over the 24-line window above each `MM_TitleSetup_Init` transition. A substring search cannot distinguish a call from prose. Replacing the guarded wrapper in `z_play.c` with:

```c
// see MM_Combo_OwlSaveExitToOoT
STOP_GAMESTATE(&this->state);
SET_NEXT_GAMESTATE(&this->state, MM_TitleSetup_Init, sizeof(TitleSetupState));
```

took the suite from **`2 failed, 10 passed`** to **`12 passed`** — green, with #532's bootstrap-authoring mechanism fully re-armed.

This is not a contrived edit. The natural way to document a TitleSetup transition you deliberately left alone is to name the guard the other ones use — which is precisely the comment that silences the lock. Given that this PR's own design *invites* prose about the guard at nearby call sites (the game-over hole, the boot-chain bail), the hole was likely to be hit by the next person to touch this code.

**Fix:** the guard is now matched against comment-stripped code (`_strip_comments` handles both `//` and `/* */`), and the `extern` prototype is excluded so a forward declaration drifting within the lookback cannot stand in for the call either. The `RSBS-TITLESETUP-EXEMPT` marker keeps searching the **raw** window — it is a comment by design; only the guard has to be real code.

Added `test_titlesetup_guard_cannot_be_satisfied_by_a_comment`, pinning all four cases directly (real call → guarded; `//` mention, block-comment mention, and `extern` prototype → NOT guarded), so a future simplification back to a raw text search fails *there* rather than silently re-opening the hole.

### Lock falsification (both locks, executed this pass)

**Lock 2 — WIRING.** Falsified twice, before and after the fix:

- *Guard deleted from `z_play.c`* → red, naming the site:
  ```
  AssertionError: expected at least the two #532-guarded transitions, found
    ['games/mm/src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope_NES.c:3768']
  FAILED test_titlesetup_transitions_are_guarded_or_marked_exempt
  FAILED test_titlesetup_invariant_sees_the_known_transitions
  2 failed, 10 passed
  ```
- *Comment bypass* → **was** green (`12 passed`), now red on both rows (`2 failed, 11 passed`).
- Restored: **13 passed**.

**Lock 1 — DECISION (runtime).** Re-ran the CF-2 inert-variant counterfactual independently (dropped `Combo_RequestGameSwitch()`, kept `return 1`; one-TU edit + incremental relink):

```
[MM] owl/pause save exit in a cross-game session: returning to OoT
[TEST] FAIL: refusing TitleSetup is not enough: the exit must request the switch
             that breaks MM's graph loop, or MM keeps running and reaches
             TitleSetup anyway
             (games/mm/2s2h/mm_unified_save_test.cpp:226)
0% tests passed, 1 tests failed out of 1
```

Verbatim reproduction of the prior pass's result. Restored and rebuilt; green again.

Also verified 6c's setup is honest: `RsbsSave_ResetSlotSessionState()` + `RsbsSave_SetActiveSlot(2)` makes the capture refuse **via the #533 latch specifically** (`slot >= 0`, so it reaches `RsbsSave_Save` → `CheckWriteAllowed` → not armed), not via the `no active slot` early-out. The log confirms it: `slot 2 NOT saved: write latched: slot was not loaded, created, or erased this session`.

### Verification

Fresh worktree at `2b96114e`, `git submodule update --init --recursive`, vcvars64 wrapper, quoted `VCPKG_ROOT`, `-j4` at low priority throughout, one build at a time.

```
cmake -S . -B build-cmake -G Ninja -DCMAKE_MAKE_PROGRAM=ninja \
      -DCMAKE_BUILD_TYPE:STRING=Release -DREDSHIP_BUILD_SHARED=ON   -> ok (52.7s)
cmake --build build-cmake -- -j4                                    -> 3015/3016, EXIT=0

ctest --test-dir build-cmake -L "^redship$"  ->  73/73 PASSED   (MMUnifiedSaveCapture #66 green)
ctest --test-dir build-cmake -L "^rando$"    ->  15/15 PASSED   (MMOwlSaveArmState #86 green)

python -m pytest tools/tests/test_repo_invariants.py -q  ->  13 passed
```

Both tiers re-run on the final restored tree after all counterfactuals; `git status --porcelain` empty (generated `build.c` / `properties.h` stamps reverted, not carried).

Confirming the prior pass's staging note, which bit here too: the `rando` tier needs `soh.o2r` + `2ship.o2r` in the build root or `RandoGenFullInit` fails with *"no soh.o2r resolvable"*. `cmake --build build-cmake --target GenerateSohOtr Generate2ShipOtr` produces both — **ROM-free**, no `ExtractAssets` needed. Staging failure, not a regression signal.

### Residuals — disclosed, not fixed

1. **Launcher-refusal livelock** (found by the prior pass; I confirmed the mechanism independently). The guarded branch skips `STOP_GAMESTATE`, and `z_play.c:874` sets `transitionTrigger = TRANS_TRIGGER_OFF` while `transitionMode` stays `TRANS_MODE_INSTANCE_RUNNING` — so the same case re-enters and re-requests **every frame**. Harmless when the switch succeeds (the graph loop breaks first), but each of `rsbs/src/main.cpp`'s four refusal branches does `Combo_ClearGameSwitchRequest(); continue;`, which resumes the still-active MM in the same parked gamestate → request/refuse livelock where vanilla had a TitleSetup fallback. All four refusals are effectively unreachable in the #532 scenario. Cheapest insurance if anyone wants it: set `transitionMode = TRANS_MODE_OFF` in the guarded branch.

2. **Lock 2 scanner scope.** It walks only `games/mm/src/**/*.c`. Every live `MM_TitleSetup_Init` transition is in that set today (verified by grep across the repo), so there is no gap now — but a transition added from a `games/mm/2s2h/*.cpp` TU would be invisible to the lock.

3. **Game-over "don't continue"** carries the identical mechanism and is a known live hole, correctly excluded as a death path and marked `RSBS-TITLESETUP-EXEMPT(#532)` so the lock names it rather than letting it hide. Wants its own issue.

4. **One-game ordering.** An owl save means "save and quit"; this delivers "save and teleport into OoT gameplay at Hyrule Market", and it is a half-file commit. Neither is introduced here — the capture is already on main — and losing the MM half is strictly worse than landing in the wrong place, so the fix is a real improvement. Both want their own issues; already recorded as dissent on an earlier comment, not duplicated.

5. **Pre-existing:** `MM_Combo_CaptureSaveToUnifiedSlot` runs `MM_HarvestSharedResources()` *before* `RsbsSave_Save`, so a latch-refused capture still advances the in-memory shared pool for a commit that never lands (harvest/apply gate-symmetry class).

### Recommendation

**KEEP OPEN and merge when a shepherd is available.** The fix is required, composes correctly with the phase-1 machinery, and both locks are now non-vacuous and demonstrated capable of going red. Not merging per lane policy.

spencerduncan:
## Adopted, merged with main, and verified locally

Merge commit `f13b9a9a`. Main had advanced by #580, #581, #586, #587, #588; the branch is now on top of `432a859a`.

### Not superseded by phase 1 — checked, not assumed

At `origin/main`, `games/mm/src/code/z_play.c:820-822` still reads:

```c
if (gSaveContext.gameMode == GAMEMODE_OWL_SAVE) {
    STOP_GAMESTATE(&this->state);
    SET_NEXT_GAMESTATE(&this->state, MM_TitleSetup_Init, sizeof(TitleSetupState));
}
```

#586's funnel made the owl save **commit** correctly. It did nothing about the bootstrap authored immediately afterward. If anything it raises the stakes: the funnel is what puts real bytes in Tier-3, and the bootstrap is what eats them. Reading `z_message.c` end to end confirms the two are sequential, not competing — `MSGMODE_OWL_SAVE_0` marshals (funnel commits inside `func_8014546C`), `MSGMODE_OWL_SAVE_1` sets `GAMEMODE_OWL_SAVE` and starts the transition, and only then does `Play_UpdateTransition` reach this branch. Commit, then exit.

### Conflict resolution

Both conflicts came from #586 (`0ded5df3`). Both intents survive; **`git diff origin/main HEAD` shows zero removals in either file**, i.e. nothing of #586's was dropped to make room.

**`z_kaleido_scope_NES.c`** — #586 retired the local `extern int MM_Combo_CaptureSaveToUnifiedSlot(void)` (the capture now rides the marshal tail in `z_sram_NES.c`) and deleted the comment block explaining why decomp TUs declare `Combo_*` entry points locally. This PR had appended its own `extern int MM_Combo_OwlSaveExitToOoT(void)` inside that block. Resolution keeps the retirement, keeps the #532 prototype, and re-states the include-path rationale the prototype now solely owns. Everything else is untouched on both sides: #586's `func_8014546C` ordering and its `Sram_FileNumHasFlashSlot` guard on the game-over flash write (the real OOB fix), and this PR's guard on the save-and-quit transition plus the `RSBS-TITLESETUP-EXEMPT` marker naming the game-over "don't continue" hole.

**`mm_unified_save_test.cpp`** — a pure numbering collision: both sides appended a block numbered "6". #586's route checks (6–10) are kept **verbatim and in order** — they are sequenced against specific slot arming, poison values and `commitGeneration` deltas, and reordering them would be a silent re-verification of somebody else's lock. This PR's exit-policy rows move after them and are renumbered 11a/11b/11c. **No row was dropped or weakened: `git diff` against the pre-merge branch removes 0 `USAVE_ASSERT` lines**, only comment prose. The move strengthens 11c if anything — it still finds slot 2 holding the healthy file checks 3-5 wrote, and the `RsbsSave_ResetSlotSessionState()` before it is still what makes the capture latch-refused.

### Both tiers, verbatim

These locks had never been observed passing on this branch (the prior lane hit disk exhaustion, and no CI job runs the ctest tiers). They pass. Full ROM-staged Ninja/Release build, capped at `-j4`.

```
100% tests passed out of 73

Label Time Summary:
redship    =  22.81 sec*proc (73 tests)

Total Test time (real) =  22.90 sec
```

```
100% tests passed out of 18

Label Time Summary:
rando    =  44.02 sec*proc (18 tests)

Total Test time (real) =  44.05 sec
```

Plus `tools/tests/`: **159 passed, 17 skipped** — this is where the source-text wiring lock lives (it runs in `python-tests.yml`, not ctest).

`MMUnifiedSaveCapture` is row 66 and now carries **both** lock sets: #586's checks 6–10 and this PR's 11a/11b/11c. No decision row failed, and none was weakened to get there.

### Falsifiability, executed three ways

**1. Revert the `z_play.c` guard** (back to main's unwrapped two lines). The source-text lock goes red and names the line:

```
E  AssertionError: MM_TitleSetup_Init transitions with neither a MM_Combo_OwlSaveExitToOoT guard nor a RSBS-TITLESETUP-EXEMPT marker (#532 -- these author a vanilla bootstrap over a live cross-game session): ['games/mm/src/code/z_play.c:833']
E  assert not ['games/mm/src/code/z_play.c:833']
```

`test_titlesetup_invariant_sees_the_known_transitions` goes red alongside it (`found ['...z_kaleido_scope_NES.c:3762']`, `assert 1 >= 2`) — the anti-vacuity check doing its job. Restored: `13 passed`.

Worth stating plainly: **no ctest row can observe this revert.** The wiring is a call-site decision; the C locks see only the helper. That is exactly what the PR body claims, and it is why the Python invariant exists. So the ctest tier got its own two counterfactuals:

**2. Gut the helper** (`return 0` unconditionally — the fix absent). Row 66 goes red at 11b, and *only* there:

```
[TEST] FAIL: a cross-game MM owl-save exit must refuse MM's TitleSetup chain (#532) (games/mm/2s2h/mm_unified_save_test.cpp:616)
```

The run log shows #586's route checks still passing underneath it — commit generations 3→6, the #533 latch refusal, the #570 identity refusal — so the perturbation is precisely targeted at #532 and does not disturb #530's locks.

**3. Drop the `Context_HasFrozenState` gate** (always switch). Row 66 goes red at 11a instead, i.e. the "must not hijack standalone 2ship" guarantee is real and not decorative:

```
[TEST] FAIL: a standalone MM session must keep the vanilla TitleSetup exit (#532 gate) (games/mm/2s2h/mm_unified_save_test.cpp:600)
```

Restored, rebuilt, and both tiers re-run green on the final tree — the numbers quoted above are from that final run, not from before the counterfactuals.

### Housekeeping

- `libultraship` moves to `477633a7`, which is exactly `origin/main`'s pointer (#587's bump), not local drift.
- Generated stamps (`games/{oot,mm}/src/boot/build.c`, `games/oot/properties.h`, `games/mm/windows/properties.h`) reverted before commit.
- Zero formatting delta: clang-format reports an identical replacement count on `origin/main` and on the merged tree for both touched decomp files, so every remaining replacement is pre-existing 14-vs-19 version drift.

Not merging — leaving that to the shepherd.

Issue #395, opened 2026-07-20
CONFIRMED: MM writes 92 bytes past a 4-byte GameInteractor allocation (integration-test paths; tree-wide layout landmine)


spencerduncan:
**Machine-verified.** Every MM hook registration writes ~92 bytes past the end of a 4-byte heap allocation, on every MM boot in the shipping binary. This is a strong candidate root cause for the long-open Fault A (`0xC0000374` heap corruption on exit).

This is the GameInteractor half of #383, escalated from "layout mismatch" to "confirmed live OOB write" by direct compiler probes rather than static reasoning.

## How it was verified

Both games' real translation units were compiled with their exact production flags (extracted from `build-cmake/build.ninja`, MSVC, Release), with probes inserted and then reverted:

**OoT's allocation site** — `games/oot/soh/OTRGlobals.cpp:1709`, immediately before `GameInteractor::Instance = new GameInteractor();`:

```cpp
static_assert(sizeof(GameInteractor) == 4, "...");   // COMPILES CLEAN -> it IS 4
```

**MM's translation unit** — `games/mm/2s2h/GameExports_SingleExe.cpp`:

```cpp
template <int N> struct RsbsSizeProbe;
static RsbsSizeProbe<(int)sizeof(GameInteractor)> p;
// error C2079: uses undefined struct 'RsbsSizeProbe<104>'

template <int N> struct RsbsOffProbe;
static RsbsOffProbe<(int)offsetof(GameInteractor, nextHookId)> q;
// error C2079: uses undefined struct 'RsbsOffProbe<96>'
```

| | OoT | MM |
|---|---|---|
| `sizeof(GameInteractor)` | **4** | **104** |
| `offsetof(nextHookId)` | 0 | **96** |

## The write

`games/mm/2s2h/GameExports_SingleExe.cpp:342, :373, :414, :497` call
`GameInteractor::Instance->RegisterGameHook<GameInteractor::OnGameStateMainStart>(...)`.
`RegisterGameHook` reads and writes `nextHookId`.

MM's code computes that member at **offset 96** of an object OoT allocated as **4 bytes**. So each registration reads and writes bytes 96–99 of a 4-byte block — **92 bytes past the end**, straight into glibc/CRT heap metadata of the neighbouring chunk.

Those four call sites are in `--integration-test` arming paths, so the *frequency* depends on mode, but they are compiled into the shipping binary and MM's boot path reaches them.

## Why it stayed hidden

[...]

PR #550, opened 2026-07-29
Tag-scope MM's GameInteractor hook types so the two ports' hook registries cannot COMDAT-fold (#470)

### LINKED ISSUE #470: LATENT: GameInteractor::RegisteredGameHooks<H> COMDAT-folds OoT and MM hook registries with divergent std::function signatures
Found by the systematic cross-tree ODR audit in #468 (classification B2, LATENT). Cross-referenced from #427; the arming flips are tracked in #392.

## The hazard

Both ports define a global-namespace `class GameInteractor` containing a byte-identical nested template:

```cpp
template <typename H> struct RegisteredGameHooks {
    inline static std::unordered_map<HOOK_ID, typename H::fn> functions;
    inline static std::unordered_map<int32_t, std::unordered_map<HOOK_ID, typename H::fn>> functionsForID;
    inline static std::unordered_map<uintptr_t, std::unordered_map<HOOK_ID, typename H::fn>> functionsForPtr;
    inline static std::unordered_map<HOOK_ID, std::pair<typename H::filter, typename H::fn>> functionsForFilter;
};
```

- OoT: `games/oot/soh/Enhancements/game-interactor/GameInteractor.h:231`
- MM: `games/mm/2s2h/GameInteractor/GameInteractor.h:184`

`GameInteractor::RegisteredGameHooks<GameInteractor::OnSceneInit>::functions` mangles on the **enclosing class name and the hook-type name only** — the `fn` typedef *inside* `H` is not part of the mangled name. So OoT's and MM's registries for any same-named hook type COMDAT-fold into **one** object, whose `std::function` payloads have **different call signatures** on the two sides.

This is invisible to every existing gate:

- **#375 symbol-collision gate** — these are `inline static` template members, i.e. weak/COMDAT symbols, explicitly out of that script's documented scope ("templates and inline functions from shared headers legitimately COMDAT-fold").
- **Layout comparison** — the two type *names* are identical by design; the divergence lives in a template argument.
- **`mm_gi_hook_guard.h`** — poisons only `Register*`. `Execute*`, `Unregister*`, and `GetHookData` are deliberately permitted.

## The divergent hook types

16 hook-type names are shared between the two `GameInteractor_HookTable.h` files; **14 have divergent signatures**:

| Hook | OoT signature | MM signature |
|---|---|---|
| `OnActorInit` / `OnActorUpdate` / `OnActorKill` / `OnActorDestroy` | `(void* actor)` | `(Actor* actor)` |
| `ShouldActorInit` / `ShouldActorUpdate` | `(void* actor, bool* result)` | `(Actor* actor, bool* should)` |
| `OnSceneInit` | `(int16_t sceneNum)` | `(s8 sceneId, s8 spawnNum)` |
| `OnKaleidoUpdate` | `()` | `(PauseContext* pauseCtx)` |
| `OnFlagSet` / `OnFlagUnset` | `(int16_t, int16_t)` | `(FlagType, u32)` |
| `OnSceneFlagSet` / `OnSceneFlagUnset` | `(int16_t, int16_t, int16_t)` | `(s16, FlagType, u32)` |
| `OnOpenText` | `(uint16_t*, bool*)` | `(u16*, bool*)` |
| `OnSeqPlayerInit` | `(int32_t, int32_t)` | `(s32, s32)` |

The only two same-named hooks that agree are `OnGameStateMainStart` and `OnPlayDestroy`.

`OnSceneInit` and `OnKaleidoUpdate` differ in **arity** — the sharpest case. A handler registered as `std::function<void(s8,s8)>` and invoked through a `std::function<void(int16_t)>` interface calls the wrong invoker with garbage arguments. The pointer-type cases (`void*` vs `Actor*`) are less immediately fatal but equally undefined.

## Why it is not armed today

MM instantiates a registry only by naming `Register*` / `Execute*` / `Unregister*` / `GetHookData` for a given hook type.

- `Register*` is poisoned by `mm_gi_hook_guard.h` in `2ship_port`, `2ship_src`, `2ship_rando`, `2ship_rando_ui`.
- The only `Execute*` instantiations in linked MM TUs are `games/mm/2s2h/GameExports_SingleExe.cpp:1595` and `:1599` — `OnRoomInit` and `AfterRoomSceneCommands`. Both are **MM-only names, absent from OoT's 81-entry table**, so neither folds.

Verified: `grep -c "OnRoomInit\|AfterRoomSceneCommands" games/oot/soh/Enhancements/game-interactor/GameInteractor_HookTable.h` returns 0.

## What arms it

Either of these, with **no link error and no CI signal**:

1. **The `2ship_enh` flip (#392).** `2ship_enh` is exempt from the hook guard and holds ~21 direct `Register*` sites. Un-eliding it introduces MM instantiations of same-named registries wholesale.
2. **A single new `ExecuteHooks<...>` wrapper in `GameExports_SingleExe.cpp`** naming any of the 14 hooks above. This is a natural next step for MM scene/flag work — `ExecuteHooks<GameInteractor::OnSceneInit>` is exactly the kind of line someone would add — and **nothing currently blocks it**. The guard permits `Execute*` by design, and its stated rationale ("only touch the inline-static per-hook-type maps (no instance state), so they stay usable") is correct about instance state and silent about registry folding.

Risk 2 is the more likely one, because it looks safe and is explicitly sanctioned by the guard's comment.

## Proposed fix

Two options, roughly in order of preference:

**A. Widen the guard to cover same-named hook types.** Extend `games/mm/include/mm_gi_hook_guard.h` to also poison `ExecuteHooks` / `ExecuteHooksForID` / `ExecuteHooksForPtr` / `ExecuteHooksForFilter`, and give `GameExports_SingleExe.cpp` a narrow, explicitly-`#undef`-ed escape hatch for the two MM-only hooks it legitimately dispatches (the `mm_culling_test.cpp` `#undef` precedent). This makes the dangerous case fail at compile time and forces any new dispatch to be a deliberate, reviewed act.

**B. Disambiguate the registry symbol.** Give MM's `RegisteredGameHooks` an MM-distinct name or nest MM's `GameInteractor` in `S2H`. Cleaner in principle, but #395 records why namespacing MM's `GameInteractor` wholesale does not transfer: MM's only allocator (`BenPort.cpp`) is an excluded TU, so an `S2H`-namespaced `Instance` would never be allocated. A narrower rename of just the nested registry template may sidestep that.

Either way, a companion runtime lock in the shape of `mm_gi_shim_test.cpp` — asserting the two ports' same-named hook signatures still diverge, so the fault class is proven live rather than assumed — would keep the reasoning honest, matching the "layout premise, machine-verified" pattern already used there.

## Acceptance

- Adding `ExecuteHooks<GameInteractor::OnSceneInit>` to an MM TU fails to build (red-then-green demonstrated), **or** the registry symbols are provably distinct across the two archives.
- The `2ship_enh` flip in #392 cannot land without this resolved first.

spencerduncan:
## Summary

Closes the latent ODR hazard in #470: OoT's and MM's `GameInteractor::RegisteredGameHooks<H>` / `HooksToUnregister<H>` inline statics mangle on the enclosing class name plus the hook-**type name** only, so a same-named hook instantiated by both trees folds into one object whose `std::function` payloads have different call signatures.

In single-exe builds MM's hook types are now defined inside a nested `GameInteractor::MM_HookTypes` tag scope, with X-macro-generated `using` aliases re-exporting every name at the old `GameInteractor::<Hook>` spelling. Every hook-keyed MM mangled name therefore embeds the tag and can never collide with OoT's.

## Mechanism / motivation

`H::fn` — the divergent part — never enters an Itanium variable mangling, so `_ZN14GameInteractor19RegisteredGameHooksINS_11OnSceneInitEE9functionsE` is emitted identically by both trees and GNU ld silently keeps one. 14 of the 16 hook names shared by the two `GameInteractor_HookTable.h` files have divergent payloads; `OnSceneInit` (OoT `void(int16_t)` vs MM `void(s8, s8)`) and `OnKaleidoUpdate` differ in **arity**. MSVC embeds the variable's type in static-data-member manglings, so Windows only folds the four identical-payload names — Linux/macOS are where the fault bites, and where the locks have teeth.

Route B from the issue, narrowed: the class name is untouched, so the `Instance` cross-binding and the #395 shim contract are unaffected (the #395 objection to namespacing MM's `GameInteractor` wholesale does not apply). Tagging the hook types rather than renaming only the registry struct also kills the sibling hazard where `ExecuteHooks<H, Args...>` instantiations with mangle-identical `Args` (`OnSeqPlayerInit(s32,s32)`, `OnOpenText(u16*,bool*)`) would still fold cross-tree with divergent bodies. `S2H::GameHooks` is untouched; non-single-exe builds see pristine upstream (`#if defined(RSBS_SINGLE_EXECUTABLE)` on both added blocks, the upstream `DEFINE_HOOK` block itself textually unchanged).

## What changed

- `games/mm/2s2h/GameInteractor/GameInteractor.h` — `struct MM_HookTypes { ... }` around the `DEFINE_HOOK` expansion, then a second table pass emitting `using <Hook> = MM_HookTypes::<Hook>;` at class scope. All ~650 `COND_HOOK` / `COND_ID_HOOK` / `COND_VB_SHOULD` sites, `HookDebugger.cpp`, and the `GameExports_SingleExe.cpp` executors compile textually unchanged.
- `games/mm/2s2h/GameExports_SingleExe.cpp`, `games/oot/soh/Enhancements/game-interactor/GameInteractor_Hooks.cpp` — per-tree probes exporting the addresses of that tree's `RegisteredGameHooks<OnSceneInit>::functions` / `HooksToUnregister<OnSceneInit>::hooks`, each with a `static_assert` pinning its own side's payload shape so the divergence premise is machine-verified rather than assumed.
- `games/mm/2s2h/mm_gi_shim_test.cpp` — new section 1b in the existing `mm-gi-shim` row.
- `.github/scripts/check-symbol-collisions.sh` — new #470 section (see below).
- `games/mm/include/mm_gi_hook_guard.h`, `games/mm/include/mm_game_hooks.h`, MM `GameInteractor.h` macro-redirection header comment — the guard's "`Execute*` stays usable" rationale now documents *why* that permission is fold-safe, and the two comments that still asserted the fold hazard was live are corrected. `mm_game_hooks.h` in particular was the canonical narrative for why `S2H::GameHooks` exists; after this change the #395 out-of-bounds write on the shared 4-byte instance is the standing reason on its own, and the file now says so instead of leaving a future reader to re-derive a stale folding argument.

## Lock note

Two falsifiable locks, both red if the tag scope is reverted:

1. **Runtime** — `redship --test mm-gi-shim` (CTest `MMGIShim`) asserts the two trees' `RegisteredGameHooks<OnSceneInit>::functions` and `HooksToUnregister<OnSceneInit>::hooks` are at different addresses. The getters are opaque cross-TU calls returning addresses of writable data, so equal addresses can only mean a fold. MM's probe TU instantiates the same-named-hook registry **on purpose** — that is precisely the arming scenario the issue describes.
2. **Archive-level** — `.github/scripts/check-symbol-collisions.sh` now intersects the `_ZN14GameInteractor{19RegisteredGameHooks,17HooksToUnregister}I…` symbols of `libsoh_*.a` against `lib2ship_*.a` (weak/unique included — deliberately outside the strong-symbol gate's documented scope) and fails on any intersection. It refuses to pass vacuously if either side collects zero such symbols, and the deliberate MM probe instantiation guarantees a real symbol exists to fold on revert.

Teeth are on the Linux legs: MSVC never folds this particular pair, so a revert must not be "verified" on Windows alone. The header, the test, and the gate all say this explicitly.

Not built locally (submodules uninitialised in the review worktree); correctness rests on reading the real headers and the two hook tables, plus CI. Reviewed adversarially against the real declarations: hook-type access is `public` throughout `GameInteractor`, `RemoveAllQueuedHooks`'s earlier `ProcessUnregisteredHooks<name>()` expansion resolves through the aliases because member-function bodies are complete-class contexts, the two `);`-terminated table entries yield a permitted empty-declaration at class scope, `GameInteractor_HookTable.h` has no include guard (it is already double-included upstream), `RSBS_SINGLE_EXECUTABLE` is set on all five MM targets so no MM TU can see an untagged variant, and both probe TUs land in archives the gate's globs cover (`libsoh_enh.a`, `lib2ship_port.a`).

Closes #470

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8738619306.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8736658111.zip)

### DIFF
--- a/.github/scripts/check-symbol-collisions.sh
+++ b/.github/scripts/check-symbol-collisions.sh
@@ -139,3 +139,57 @@ if [ -s "$TMP_DIR/new" ]; then
 fi
 
 echo "OK: soh_*/2ship_* symbol intersection matches baseline ($COUNT symbol(s), all tolerated)"
+
+# --------------------------------------------------------------------------
+# #470: GameInteractor hook-registry statics. These are WEAK/unique symbols
+# and so exempt from the strong-symbol gate above — but they are the one
+# vague-linkage family where cross-tree folding is NOT legitimate: both ports
+# define global-namespace GameInteractor::RegisteredGameHooks<H> /
+# HooksToUnregister<H>, mangled on the enclosing class + hook-type NAME only,
+# while H::fn (the std::function payload, divergent for 14 of the 16 shared
+# hook names) never enters the symbol. A fold merges two incompatible
+# registries into one object. MM's hook types are tag-scoped under
+# GameInteractor::MM_HookTypes (games/mm/2s2h/GameInteractor/GameInteractor.h)
+# precisely so no such symbol can be emitted by both trees; this probe fails
+# the build if one ever is. Anti-vacuity: GameExports_SingleExe.cpp
+# instantiates MM's OnSceneInit registry on purpose (a hook name OoT also
+# defines), so a tag revert produces a real intersection here instead of an
+# empty-set pass.
+collect_gi_registry() {
+    local out="$1"
+    shift
+    : > "$out.raw"
+    local lib
+    for lib in "$@"; do
+        [ -f "$lib" ] || continue
+        # Keep ALL defined external symbols (weak W/V and GNU-unique u
+        # included — inline statics land there); match the two registry
+        # templates by their Itanium-mangled scope prefix.
+        nm --defined-only --extern-only "$lib" 2>/dev/null |
+            awk 'NF == 3 { print $3 }' |
+            grep -E '^_ZN14GameInteractor(19RegisteredGameHooks|17HooksToUnregister)I' >> "$out.raw" || true
+    done
+    sort -u "$out.raw" > "$out"
+    if [ ! -s "$out" ]; then
+        echo "error: zero GameInteractor hook-registry symbols in: $* — the #470 probe instantiations" >&2
+        echo "(OoT: GameInteractor_Hooks.cpp, MM: GameExports_SingleExe.cpp) or the mangling scheme moved?" >&2
+        echo "Refusing to pass vacuously; fix the collection here rather than losing the guard." >&2
+        exit 2
+    fi
+}
+
+collect_gi_registry "$TMP_DIR/soh_gireg"   "$BUILD_DIR"/games/oot/libsoh_*.a
+collect_gi_registry "$TMP_DIR/2ship_gireg" "$BUILD_DIR"/games/mm/lib2ship_*.a
+
+comm -12 "$TMP_DIR/soh_gireg" "$TMP_DIR/2ship_gireg" > "$TMP_DIR/gireg_folded"
+
+if [ -s "$TMP_DIR/gireg_folded" ]; then
+    echo "FAIL: GameInteractor hook-registry symbol(s) emitted by BOTH soh_* and 2ship_* archives:" >&2
+    demangle < "$TMP_DIR/gireg_folded" | sed 's/^/  /' >&2
+    echo "The linker will keep ONE of each — merging two hook registries whose std::function" >&2
+    echo "payloads have different call signatures (#470). MM hook types must stay tag-scoped" >&2
+    echo "under GameInteractor::MM_HookTypes (games/mm/2s2h/GameInteractor/GameInteractor.h)." >&2
+    exit 1
+fi
+
+echo "OK: GameInteractor hook-registry symbols are disjoint across the two trees (#470)"
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -19,7 +19,8 @@
 
 #ifdef RSBS_SINGLE_EXECUTABLE
 
-#include <cstddef> // offsetof for the #395 layout facts; ptrdiff_t
+#include <cstddef>     // offsetof for the #395 layout facts; ptrdiff_t
+#include <type_traits> // std::is_same_v for the #470 payload-divergence premise
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -689,6 +690,29 @@ extern "C" size_t MM_GI_InstanceSize(void) {
 
 extern "C" size_t MM_GI_NextHookIdOffset(void) {
     return offsetof(GameInteractor, nextHookId);
+}
+
+// #470 registry-identity probes — MM's view of the OnSceneInit hook registry.
+// Naming RegisteredGameHooks<OnSceneInit> here instantiates, from a linked MM
+// TU, exactly the symbol #470's arming scenario would create: OnSceneInit is
+// a hook NAME both ports' tables define, with divergent payloads (OoT
+// std::function<void(int16_t)>, MM the two-arg form the static_assert pins).
+// The MM_HookTypes tag scope (2s2h/GameInteractor/GameInteractor.h) keeps
+// this instantiation's mangled name disjoint from OoT's, so it is safe to
+// emit — and it is emitted ON PURPOSE: the mm-gi-shim lock asserts these
+// addresses differ from the OoT twins' (GameInteractor_Hooks.cpp), and the
+// check-symbol-collisions.sh #470 probe needs a same-named-hook registry in
+// the 2ship archives so a tag revert folds a real symbol instead of passing
+// vacuously.
+static_assert(std::is_same_v<GameInteractor::OnSceneInit::fn, std::function<void(s8, s8)>>,
+              "MM's OnSceneInit payload changed shape — revisit the #470 registry-identity lock");
+
+extern "C" void* MM_GI_OnSceneInitRegistryAddr(void) {
+    return &GameInteractor::RegisteredGameHooks<GameInteractor::OnSceneInit>::functions;
+}
+
+extern "C" void* MM_GI_OnSceneInitUnregQueueAddr(void) {
+    return &GameInteractor::HooksToUnregister<GameInteractor::OnSceneInit>::hooks;
 }
 
 // ============================================================================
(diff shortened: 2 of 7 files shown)


spencerduncan:
**CI failures diagnosed and fixed** (48789e96).

**What failed:** build-linux, build-windows, and netplay-relay all died compiling hundreds of MM TUs with stray-token errors (`stray '#'`, `missing terminating ' character`, `extended character —`) pointing at `games/mm/include/mm_gi_hook_guard.h`, cascading into whatever each TU included next (spdlog, ucrt `assert.h`, `ObjectExtension.h`).

**Root cause:** one character pair. The new #470 fold-safety paragraph in the guard header's doc comment wrote the member family as `Execute*/GetHookData` — the `*/` inside that phrase terminates the `/** ... */` block at line 21, so lines 21–50 of prose (apostrophes, em dashes, `#395`/`#470`, backticks) lex as code. The header is force-included into every MM C++ TU, hence the blast radius; all three jobs share this single cause (the Windows log blames the same header ~300 times).

**Fix:** spell it `Execute* or GetHookData`. Comment-only — the tag scope, both probe TUs, the mm-gi-shim 1b lock, and the archive gate are byte-identical.

Also re-verified while in there: no other added line in the PR carries a mid-prose `*/` inside a block comment (the one in `GameInteractor.h` is in a `//` comment), the two payload static_asserts match the real hook tables (MM `OnSceneInit(s8, s8)`, OoT `OnSceneInit(int16_t)`), and no CI error was ever attributed to the functional changes themselves — `GameInteractor.h`'s tag scope is force-included *before* the guard and parsed clean under both GCC 11 and MSVC.

Issue #560, opened 2026-07-30
Menu seed generation races the render thread over an unlocked zip_t, and a null pathway resource turns the race into a crash


spencerduncan:
Operator-reported crash, 2026-07-29: creating a new rando file on the f9b525bd artifact AV'd (0xC0000005) 65ms into "Shuffling Shop Items", preceded by two `ResourceLoader` failures reading valid archive entries as all zeros ("type 0x0, Format 0, Version 0": `scenes/nonmq/HIDAN_scene/HIDAN_scenePathwayList_000450` and `textures/nes_font_static/gMsgChar2EFullStopTex`). Diagnosed by a 4-lane multi-agent sweep; all lanes converge on one mechanism.

## Mechanism

libultraship's archive layer reads a **single shared `zip_t*` with no synchronization anywhere** (`O2rArchive::LoadFile`, `libultraship/src/ship/resource/archive/O2rArchive.cpp:22-67`; grep for mutex/lock_guard over `src/ship/resource/archive/*.cpp` returns zero hits; `ResourceManager::mMutex` guards only the resource cache map). libzip's `zip_t` is documented not-thread-safe. Two threads read it concurrently during menu-triggered seed generation:

1. **Gen side:** `GenerateRandomizer()` spawns a worker (`randomizer.cpp:3510-3516`). The session's FIRST generation loads full dungeon scenes during fill — `fill.cpp:1299 AssumedFill` → `Logic::SmallKeys` → `GetDungeonSmallKeyDoors` (`logic.cpp:2390-2397`) loads `scenes/nonmq/HIDAN_scene/HIDAN_scene` (once-per-process memo at `logic.cpp:2374-2379`). The scene factory sub-loads the pathway list (`soh/resource/importer/scenecommand/SetPathwaysFactory.cpp:18-21`), executing on a resource-pool worker.
2. **Render side:** the file-select menu draws "Generating..." every frame; the Fast3D interpreter resolves cold glyph textures via a direct `LoadResourceProcess` on the render thread (`libultraship/src/fast/interpreter.cpp:3439-3440`); `gMsgChar2EFullStopTex` is the '.' glyph (`z_kanfont.c:29`).

Concurrent `zip_fread` on the shared handle corrupts/zeroes both buffers → `ReadResourceInitDataBinary` sees ByteOrder/Type/Version = 0 → the exact logged error pair, twice, 0ms apart → both loads return nullptr.

[...]

Issue #557, opened 2026-07-30
IntGameplayRoundtripSoak crashes on MM's second resume: access violation at cycle-2 mm-play


spencerduncan:
`IntGameplayRoundtripSoak` (`redship --integration-test int-gameplay-roundtrip` with `RSBS_GP_CYCLES=3`) never completes a second round trip. It dies with a Windows access violation immediately after the cycle-2 `mm-stabilize -> mm-play` transition — that is, on **MM's second `Game_Resume`** of the process.

Found while doing ROM-staged local verification of the OoT arrival frame-counter reset (PR for `claude/eager-shannon-upc6wh`). **Not caused by that change** — baseline evidence below.

## Symptom

The single-cycle row `IntGameplayRoundtrip` passes consistently (~38–108 s). The soak row fails at ~24 s, always in the same place:

```
[GP-TEST] MM save continuity verified (arrival 2: rupee sentinel 101)
[GP-TEST] MM South Clock Town stable (entrance=0xD800); starting gameplay window

=== Cross-game entrance detected ===
Target: mm entrance 0xD800

=== Switching to Majora's Mask ===
[RSBS] Archive ready: ./mm.o2r
[RSBS] Archive ready: .../build-cmake/2ship.o2r
Starting Majora's Mask... (Press F10 to switch games)
File Name scenes/nonmq/Z2_CLOCKTOWER/Z2_CLOCKTOWER_room_00
[GP-TEST] phase: mm-stabilize -> mm-play

[CRASH] Windows exception: 0xc0000005
[CRASH] addr=0x7ff74eacb4c2 exe_base=0x7ff74d3f0000 exe_offset=0x16db4c2
[CRASH] access violation: READ at 0x178fd2fb6be
[GP-TEST] state (signal): phase=mm-play cycles=1/3 frames/phase=120 boot=0x01D1 warp=0x00B1 exit=0x0033
```

Notable: MM's **first** entry runs its full 120-frame gameplay window and returns to OoT cleanly, and cycle 2's MM arrival gets far enough to verify save continuity (`rupee sentinel 101`) and report South Clock Town stable. The fault lands once the live gameplay window actually starts ticking.

## Signature: wild pointer, fixed code site

Across three runs the **faulting code offset is stable** while the **accessed address is garbage and different every time**, and the operation alternates READ/WRITE:

| Run | `exe_offset` | Access |
|---|---|---|
| 1 | `0x16dc62e` | WRITE at `0x29fc74acde2` |
[...]

PR #561, opened 2026-07-30
Null-guard the resource consumers that turned a failed load into an AV, in both games (#560)

spencerduncan:
Part of #560 (consumer hardening; root fix is the archive mutex).

#560's race only became a crash because consumers dereferenced unchecked resource
loads. This makes any resource-load failure — from that race or anything else —
degrade to a logged failure instead of an access violation, on both halves of the
combo.

## Primary crash site

`games/oot/soh/resource/importer/scenecommand/SetPathwaysFactory.cpp:21` called
`path->GetPointer()` on an unchecked `LoadResourceProcess` result. Both the binary
and XML readers now log the pathway name plus the owning scene at error level, skip
the entry, and recompute `numPaths` from what actually loaded.

Skip rather than fail the whole command, deliberately: `ParseSceneCommand`
(`SceneFactory.cpp:92-97`) pushes a null into `scene->commands` and continues, so
returning null would relocate the deref into the lists that get walked later. It
also matters that every `reader->ReadUInt32()`/`ReadString()` in the loop stays
consumed — bailing mid-loop would desync the binary reader for the scene's
remaining commands.

## Factory audit (item 2)

All 40 factory `.cpp` files under `games/oot/soh/resource/importer/` were checked.
Nine perform sub-loads; the other 31 have none.

| Factory | Sub-load | Verdict |
| --- | --- | --- |
| `scenecommand/SetPathwaysFactory.cpp` | `LoadResourceProcess` ×2 (binary + XML) | **fixed** — log + skip entry, recompute `numPaths` |
| `AnimationFactory.cpp` | `LoadResourceProcess` (link-animation segment) | **fixed** — fail the animation |
| `AudioSoundFontFactory.cpp` | `LoadResourceProcess` ×12 | **fixed** at 5 sites (3 instrument note-sounds, sfx table, base font for a patch); the other 7 already used `res ? res->GetRawPointer() : nullptr` |
| `AudioSampleFactory.cpp` | `ArchiveManager::LoadFile` | **fixed** — plus a missing-`Path`-attribute check, since `LoadFile` takes a `std::string` |
| `AudioSequenceFactory.cpp` | `ArchiveManager::LoadFile` | **fixed** — plus the `!streamed && Path == nullptr` case, which dereferenced a null `seqFile` at line 359 |
| `SkeletonFactory.cpp` | `LoadResourceProcess` ×2 | already safe — `limb ? limb->GetRawPointer() : nullptr` |
| `scenecommand/SetAlternateHeadersFactory.cpp` | `LoadResourceProcess` ×2 | already safe — pushes a possibly-null header; consumers null-check |
| `scenecommand/SetCollisionHeaderFactory.cpp` | `LoadResourceProcess` ×2 | already safe — stores unchecked, but `SetCollisionHeader::GetPointer` null-checks `collisionHeader` |
| `scenecommand/SetCutscenesFactory.cpp` | `LoadResourceProcess` ×2 | already safe — same shape, `SetCutscenes::GetPointer` null-checks |

The audit's real payoff was the last two rows: they are safe **only** because the
accessor guards, and four consumers went around it.

- `z_scene_otr.cpp` `OoT_Scene_CommandPathList` — `GetPointer()[0]` is
  `paths.data()[0]`, a null deref on an empty list. Now leaves `setupPathList`
  null, which is what a scene with no `SetPathways` command produces anyway
  (`z_play.c:2106`).
- `z_scene_otr.cpp` `Scene_CommandCutsceneData` — reached the data via
  `->cutscene->commands.data()`, going around the guarded `GetPointer()`. Now uses
  the accessor.
- `logic.cpp` `GetDungeonSmallKeyDoors` — walked `scene->commands` reading
  `command->cmdId` unchecked, on the exact scene load in #560's generation chain.
  The engine's own `OTRScene_ExecuteCommands` already skips nulls; this walk now
  does too.
- `GraveHoleJumps.cpp` — scene, command, `sceneCmd`, and `GetRawPointer()` all
  unchecked.

## MM half — same class, one hole worse

The audit was repeated on `games/mm/2s2h/`.

| Site | Verdict |
| --- | --- |
| `z_scene_2SH.cpp` `MM_OTRScene_ExecuteCommands` | **fixed** — read `sceneCmd->cmdId` straight through the null entries `ParseSceneCommand` pushes (`SceneFactory.cpp:102` logs each one). OoT's equivalent loop has always had `if (sceneCmd == nullptr) continue;`; MM's did not, so **any** logged scene-command load failure was an immediate AV on the next MM scene load |
| `scenecommand/SetPathwaysFactory.cpp` | **fixed** — verbatim the same `path->GetPointer()` deref as OoT's primary crash site |
| `z_scene_2SH.cpp` `MM_Scene_CommandPathList` | **fixed** — same `GetPointer()[0]`-on-empty guard as OoT |
| `AnimationFactory.cpp` | **fixed** — same link-animation segment site as OoT's |
| `scenecommand/SetAnimatedMaterialListFactory.cpp` | **fixed** — stores null; `play->sceneMaterialAnims` is null in a scene without this command and its consumers handle that |
| `scenecommand/SetCutscenesFactory.cpp` | already safe — already had exactly this guard, which is why the pattern above matches it |
| `scenecommand/SetCollisionHeaderFactory.cpp` | already safe — accessor null-checks |
| `SkeletonFactory.cpp`, `AudioSoundFontFactory.cpp` | already safe — ternary-guarded |
| `KeyFrameFactory.cpp`, `AudioSampleFactory.cpp`, `AudioSequenceFactory.cpp`, `SetAlternateHeadersFactory.cpp` | not touched — see follow-ups |

The `MM_OTRScene_ExecuteCommands` row is the load-bearing one: it is a strictly
worse version of the hole guarded on the OoT side, and it is plausibly the same
class as #557 (wild-pointer AV on MM's second resume), which #560 already flags as
a candidate.

## fill.cpp shop shuffle (item 3)

Guarded the four `Rando::StaticData::GetLocation()` results, matching the
surrounding style. Three of them fed `GetRandomPrice`, which dereferences its
argument inside `GetPriceFromSettings`. The scrub and merchant `if`/`else` pairs
were folded into one loop each so the guard is written once; `GetRandomPrice` is
still called in exactly the same cases and in the same order, so the RNG stream is
untouched — `RandoDeterminism` and `SeedDeterminism` are the evidence.

## Failed generation no longer wedges the menu (item 4)

`logic.cpp:2396` is the only resource load in the whole generation path. A null
scene already returned the empty key-door list gracefully; the null-command walk
was the remaining AV.

The stuck-flag consequence the operator hit came from an exception escaping the
generation thread. `GenerateRandomizerImgui` **is** the thread entry
(`randomizer.cpp:3530`) and nothing above it catches, so a throw `std::terminate`'d
the process with `RandoGenerating` already persisted as 1 by the
`SaveConsoleVariablesNextFrame` at the top — leaving the menu and file-select
screen believing a generation was still in flight, and the thread never joinable.
The whole body is now inside the try, not just the `GenerateRando` call: the two
`std::stoi` loops above it parse persisted CVar strings and throw
`std::invalid_argument` on any non-numeric token, so a narrower try left the same
hole. Failure logs, marks the seed not-generated, and falls through to the existing
flag reset.

Note this catches C++ exceptions only. An access violation on Windows is SEH and
is not caught by `catch (...)`; the null guards above are what remove that.

## Verification

**Local build** — MSVC / Ninja / Release, `REDSHIP_BUILD_SHARED=ON`, rebased onto
`a1dc22ab`:

```
[491/492] Linking CXX executable redship.exe
BUILD_OK
```

**Warnings** — the full build log was grepped for every touched filename; zero
`warning`/`error` lines from any of them. (An earlier pass did surface one new
`C4267 size_t -> uint32_t` in `SetPathwaysFactory.cpp`, fixed with an explicit
`static_cast` — and the identical pre-existing one in the XML reader with it.)

**Both ROM-free ctest tiers**, on the rebased binary:

```
100% tests passed out of 65
redship    =  11.11 sec*proc (65 tests)

100% tests passed out of 14
rando    =  22.30 sec*proc (14 tests)
```

**Counterfactual** for the engine-side guard, compiled and run on this toolchain —
the factory fix alone would have moved the AV, not removed it:

```
empty vector: size=0 data()=0000000000000000 data()==nullptr: YES
one-element vector: size=1 data()=0000015001320180 data()[0]=0000000000000000
```

**clang-format** — verified with the repo's pinned 14.0.6. The two touched files on
`.github/clang-format-paths.txt` (`games/oot/soh/z_scene_otr.cpp`,
`games/mm/2s2h/z_scene_2SH.cpp`) are clean, as are the other 13. The only violation
anywhere in the touched set is `fill.cpp:402`, a pre-existing `assert` block that is
untouched and not on the allowlist.

**Remote CI**, all green at `eebcba34`: build-linux 2m57s, build-windows 13m9s,
netplay-relay 3m27s, clang-format, clang-tidy, test, generate-rsbs-otr,
odr-declaration-gate, link-check 2m55s. build-macos and create-release skipped.
`link-check` was red before the rebase; it is the #550 self-test fixture bug that
`a1dc22ab` (#563) fixed on main, not anything in this change — rebasing onto it
turned the check green with no source change here.

No submodule pointer or generated-stamp changes: the diff is 15 source files.

## Residual risk

A scene whose pathway list fails to load now gets a null `setupPathList` instead of
an AV at load time — but an actor in that scene that indexes `setupPathList[i]`
will still fault later. That is the same exposure vanilla has for a scene with no
`SetPathways` command, and it is strictly better than crashing during the load, but
it is a deferred fault rather than no fault. The archive mutex is what makes the
state unreachable.

## Follow-ups, deliberately not in this change

1. `ResourceManager` still negative-caches a transient read failure identically to a
   genuine not-found (#560 fix plan item 2) — belongs with the libultraship change.
2. The XML scene-command factories build a `std::string` from
   `XMLElement::Attribute(...)` with no null check
   (`SetPathwaysFactoryXML` "FilePath", `SetCollisionHeaderFactoryXML` /
   `SetCutscenesFactoryXML` "FileName", `SetAlternateHeadersFactoryXML` "Path") —
   missing-attribute UB, a different class from the archive race.
3. MM's `KeyFrameFactory`, `AudioSampleFactory`, `AudioSequenceFactory` and
   `SetAlternateHeadersFactory` sub-loads are enumerated above but unguarded; they
   are the audio/keyframe tail of the same sweep, not on the crash path.
4. `Rando::StaticData::GetLocation` and `Context::GetItemLocation` return
   `&table[key]` unconditionally, so they can never return null — the ~6 existing
   `loc == nullptr` guards in the tree, and the four added here, are defensive
   rather than live; the real hazard for a bad key is a pointer past the end of the
   array. Bounds-checking both accessors was implemented and then reverted: it would
   make every guard meaningful but would convert today's out-of-range reads into hard
   null derefs in the many callers that do *not* check, which works against this
   change's goal. A `NOTE` in `location_list.cpp` records this; it needs its own
   change with an audit of the unguarded callers.
5. No ROM-free ctest locks the `SetPathwaysFactory` guard, because writing one means
   pulling OoT `soh/resource/` headers into `test_runner.cpp`, which already includes
   MM's `2s2h/resource/` headers in the same TU. That OoT/MM header collision
   deserves its own change.

### DIFF
--- a/games/mm/2s2h/resource/importer/AnimationFactory.cpp
+++ b/games/mm/2s2h/resource/importer/AnimationFactory.cpp
@@ -88,6 +88,14 @@ ResourceFactoryBinaryAnimationV0::ReadResource(std::shared_ptr<Ship::File> file,
         const auto animData = std::static_pointer_cast<Animation>(
             Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(path.c_str()));
 
+        // #560: a failed sub-load returns null. Fail the whole animation rather than
+        // dereferencing it or handing the animation player an uninitialized segment.
+        // Mirrors the same site in OoT's AnimationFactory.
+        if (animData == nullptr) {
+            SPDLOG_ERROR("Animation {}: failed to load link animation segment \"{}\"", initData->Path, path);
+            return nullptr;
+        }
+
         animation->animationData.linkAnimationHeader.segment = animData->GetPointer();
     } else if (animType == AnimationType::Legacy) {
         SPDLOG_DEBUG("BEYTAH ANIMATION?!");
--- a/games/mm/2s2h/resource/importer/scenecommand/SetAnimatedMaterialListFactory.cpp
+++ b/games/mm/2s2h/resource/importer/scenecommand/SetAnimatedMaterialListFactory.cpp
@@ -3,6 +3,7 @@
 #include "2s2h/resource/type/TextureAnimation.h"
 #include <ship/Context.h>
 #include <ship/resource/ResourceManager.h>
+#include <spdlog/spdlog.h>
 
 namespace S2H {
 
@@ -16,6 +17,16 @@ SetAnimatedMaterialListFactory::ReadResource(std::shared_ptr<Ship::ResourceInitD
     std::string str = reader->ReadString();
     const auto data = std::static_pointer_cast<TextureAnimation>(
         Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(str.c_str()));
+
+    // #560: a failed sub-load returns null. play->sceneMaterialAnims is null in a
+    // scene that has no SetAnimatedMaterialList at all and its consumers handle that,
+    // so store null rather than dereferencing.
+    if (data == nullptr) {
+        SPDLOG_ERROR("SetAnimatedMaterialList: failed to load texture animation \"{}\" for {}; leaving it unset", str,
+                     initData->Path);
+        setAnimatedMat->mat = nullptr;
+        return setAnimatedMat;
+    }
 
     AnimatedMaterial* res = data->GetPointer();
     setAnimatedMat->mat = res;
--- a/games/mm/2s2h/resource/importer/scenecommand/SetPathwaysFactory.cpp
+++ b/games/mm/2s2h/resource/importer/scenecommand/SetPathwaysFactory.cpp
@@ -2,6 +2,7 @@
 #include "2s2h/resource/type/scenecommand/SetPathways.h"
 #include <ship/Context.h>
 #include <ship/resource/ResourceManager.h>
+#include <spdlog/spdlog.h>
 
 namespace S2H {
 std::shared_ptr<Ship::IResource> SetPathwaysMMFactory::ReadResource(std::shared_ptr<Ship::ResourceInitData> initData,
@@ -10,14 +11,30 @@ std::shared_ptr<Ship::IResource> SetPathwaysMMFactory::ReadResource(std::shared_
 
     ReadCommandId(setPathways, reader);
 
-    setPathways->numPaths = reader->ReadUInt32();
-    setPathways->paths.reserve(setPathways->numPaths);
-    for (uint32_t i = 0; i < setPathways->numPaths; i++) {
+    uint32_t declaredPaths = reader->ReadUInt32();
+    setPathways->paths.reserve(declaredPaths);
+    for (uint32_t i = 0; i < declaredPaths; i++) {
         std::string pathFileName = reader->ReadString();
         auto path = std::static_pointer_cast<PathMM>(
             Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(pathFileName.c_str()));
+        // #560: this sub-load can come back null — the archive layer reads one shared
+        // zip_t with no synchronization, so a concurrent read can leave the buffer
+        // zeroed and ReadResourceInitData then rejects it. Skip the entry instead of
+        // calling GetPointer() through the null. Mirrors the guard the sibling
+        // SetCutscenesFactory in this directory already has, and OoT's
+        // SetPathwaysFactory. Every read above stays consumed, so the binary reader
+        // does not desync for the scene's remaining commands.
+        if (path == nullptr) {
+            SPDLOG_ERROR("SetPathwaysMM: failed to load pathway resource \"{}\" for {}; skipping it", pathFileName,
+                         initData->Path);
+            continue;
+        }
         setPathways->paths.push_back(path->GetPointer());
     }
+
+    // Report what actually loaded, not what the file claimed.
+    setPathways->numPaths = static_cast<uint32_t>(setPathways->paths.size());
+
     return setPathways;
 }
 } // namespace S2H
--- a/games/mm/2s2h/z_scene_2SH.cpp
+++ b/games/mm/2s2h/z_scene_2SH.cpp
@@ -37,6 +37,7 @@
 #include "2s2h/resource/type/scenecommand/SetMinimapList.h"
 #include "2s2h/resource/type/scenecommand/SetMinimapChests.h"
 #include "2s2h/resource/type/scenecommand/SetCsCamera.h"
+#include <spdlog/spdlog.h>
 
 s32 MM_OTRScene_ExecuteCommands(PlayState* play, S2H::Scene* scene);
 
@@ -273,6 +274,16 @@ void MM_Scene_CommandLightList(PlayState* play, S2H::ISceneCommand* cmd) {
 
 void MM_Scene_CommandPathList(PlayState* play, S2H::ISceneCommand* cmd) {
     S2H::SetPathwaysMM* paths = (S2H::SetPathwaysMM*)cmd;
+
+    // GetPointer() is paths.data(), so indexing [0] on an empty list dereferences
+    // what data() returns for an empty vector (nullptr in practice). Same guard as
+    // OoT's OoT_Scene_CommandPathList: a scene with no usable pathway is equivalent
+    // to one with no SetPathways command at all, i.e. a null setupPathList.
+    if (paths->paths.empty()) {
+        SPDLOG_ERROR("MM scene command SetPathways has no usable pathway list; leaving setupPathList null");
+        play->setupPathList = nullptr;
+        return;
+    }
 
     play->setupPathList = (Path*)paths->GetPointer()[0];
 }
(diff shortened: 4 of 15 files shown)


spencerduncan:
## `link-check` is red, and it is not this PR — it is broken on `main`

This PR changes 11 OoT `.cpp` files and nothing else: no GameInteractor code, no `.github/` scripts or workflows, no CMake. The failing step is `bash .github/scripts/check-symbol-collisions.sh --self-test`, whose fixtures are synthesized in `/tmp` and never read repo sources. So it cannot be reacting to anything here.

### Root cause

#550 (`7bee3031`, merged 01:28) added the `collect_gi_registry` probe with an anti-vacuity guard:

```sh
if [ ! -s "$out" ]; then
    echo "error: zero GameInteractor hook-registry symbols in: $* — the #470 probe instantiations" >&2
    exit 2
fi
```

It runs unconditionally, after the intersection check, on `$BUILD_DIR/games/oot/libsoh_*.a`.

The self-test invokes the gate as a full child process against synthetic archives:

```sh
selftest_make_build_dir "$dir/clean" 0
bash "$0" "$dir/clean" --baseline "$dir/empty-baseline.txt" > "$dir/log.clean" 2>&1
selftest_expect "clean scenario (...)" 0 "$dir/log.clean" "$rc"
```

…and `selftest_build_archive` emits exactly four C symbols — the optional shared strong global, a per-side unique global, `rsbs_selftest_weak`, and the `_GLOBAL__sub_I_rsbs_selftest` alias. None can match `^_ZN14GameInteractor(19RegisteredGameHooks|17HooksToUnregister)I`.

So every self-test scenario that runs the gate to completion exits 2, and the three that expect 0 fail:

```
self-test FAIL: clean scenario (...) returned rc=2, expected 0
self-test FAIL: baselined collision scenario returned rc=2, expected 0
self-test FAIL: check against the regenerated baseline returned rc=2, expected 0
```

Each one reports `OK: soh_*/2ship_* symbol intersection matches baseline` immediately before the vacuity error — the actual gate logic passed; only the new probe's guard rejects the fixtures.

### Why no other branch has hit it yet

`link-check` runs on branch pushes, so each branch runs *its own* copy of the script. The last green run (`30508534233`, 02:28, `claude/focused-lewin-aed257`) logged `self-test: OK (all 6 scenarios behaved as expected)` — that branch was based before `7bee3031`. This PR is based on `c24980ac`, which contains it, and is the first `link-check` run to use the new script. Any branch cut from `7bee3031` or later will fail identically, `main` included.

### Fix (not taken here)

Teach `selftest_build_archive` to emit one hook-registry-shaped symbol so the fixtures satisfy the probe — e.g. a C TU with an `__asm__` name of `_ZN14GameInteractor19RegisteredGameHooksI...E` — using a *different* hook-type name on each side so the disjointness assertion stays meaningful and the clean scenario still passes. Deliberately left out of this PR: it is a CI-gate defect from another lane, and folding it in would mix an infrastructure fix into a consumer-hardening review. Flagged as a separate task.

Verification for this PR is unaffected: local MSVC/Ninja Release build clean, all 11 changed files recompiled with zero warnings, and both ROM-free ctest tiers green (65/65 `redship`, 14/14 `rando`, including `RandoDeterminism` and `SeedDeterminism`). `clang-format`, `clang-tidy`, `test`, `build-linux`, `odr-declaration-gate`, `generate-rsbs-otr` and `netplay-relay` are all green here.

spencerduncan:
Corroboration: the unrelated lane `claude/ci-otr-init-coverage-hole` (run `30513476823`, 04:19) fails `link-check` with the byte-identical error — same three scenarios, same `zero GameInteractor hook-registry symbols` message. Two independent branches touching completely different code fail the same way, which confirms the gate's self-test is broken repo-wide rather than reacting to either change.

Filed as a separate task; see the analysis above for the fix.

Issue #564, opened 2026-07-30
[Audit] One-game semantics: where existing work diverges, and the alignment plan


spencerduncan:
# One-game semantics: where existing work diverges, and the alignment plan

**The ruling** ([#500, operator comment, 2026-07-29](https://github.com/spencerduncan/redshipblueship/issues/500#issuecomment-5126492334)):

> "freezing at creation is the correct semantics for sure. keep that idea in mind with all designs. this is *one game* from a semantic standpoint."

Binding consequences recorded with it: the paired world has ONE identity fixed at ONE creation event (both option profiles, pools, direction, logic mode, seed); arrival-time divergence from creation-time identity is **corruption to detect and refuse, never a choice to honor**; merged/hybrid generation at OoT file-create is the natural implementation; #498's combo-level settings are the one game's settings, stamped into world identity at creation. Corollaries used throughout: one seed hash, one spoiler identity, one settings surface, one save identity.

Five audit lanes (settings surface, identity terms, persistence, creation event, open-bug re-triage) swept `origin/main` (anchors verified at f4ad479b and re-verified at bfde1963) plus the open fix plans. This issue merges their findings: the violations table, the one-authority persistence model, the target creation-event contract, and a sequenced alignment plan. Nothing here is a fix; it is the map of what must change and in what order.

---

## Violations table (merged and deduplicated across lanes)

Sizes: S = hours, M = days, L = a wave, XL = structural end-state.

### A. Creation is two events and the MM profile floats (root cluster)

| # | Violation | Where | Current vs one-game | Size | Issues |
|---|-----------|-------|---------------------|------|--------|
| V1 | **MM half is authored at first arrival, not at creation.** The entire MM world (profile snapshot, seed derivation, pools, fill, foreign placement, spoiler) runs the first time the player crosses to Clock Town. | `games/mm/2s2h/Rando/Foreign.cpp:87-151` (ResolvePairedProfile reads live CVars, generic copy loop :91-94), `games/mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp:119`, `games/mm/2s2h/GameExports_SingleExe.cpp:2728-2798`, creation seam `games/oot/src/code/z_sram.c:269-303` | Every option edit between OoT file-create and first crossing is honored and changes the world. One-game: the profile resolves once at the creation event; arrival becomes hydrate-or-refuse with zero generation capability. | L | #500 #498 #497 #493 #530 #531 #533 |
[...]

Issue #482, opened 2026-07-22
Audit: cross-game arrival-rehydration class — OoT tracker init-funcs blank out-of-save state (2 BROKEN); MM covered by the OnSaveLoad re-dispatch


spencerduncan:
Systematic audit of the cross-game **arrival-rehydration** bug class that has now bitten three times: #439/#447, #469/#472, and #441/#481.

## The class

A switch INTO a game (Happy Mask Shop / Clock Tower / F10) cold-boots that game's title/init chain, which **clears, resets, re-mallocs, or re-authors live in-memory derived state**. The frozen `SaveContext` is restored AFTERWARD at `*_Play_ConsumeStartupEntrance` via `Combo_ConsumeFrozenState`, **without re-running the normal load path** (`Save_LoadFile` / `LoadRandomizer`). So any live/derived state that (a) the boot-init chain clears and (b) the consume path does NOT rehydrate decays silently — invisible to save-file forensics because the on-disk save is correct; only the in-memory derived table is empty.

This issue is the durable artifact: the full enumeration of both games' arrival boot-init chains and the classification of every out-of-`gSaveContext` clear. Fixes for the BROKEN instances land in a follow-up PR that references this.

## Structural headline

**MM's consume path re-dispatches `OnSaveLoad`** (`games/mm/src/code/z_play.c:2425`, the #447 fix), which rebuilds the *entire* `OnSaveLoad`-driven derived-table set — so the whole MM chain is structurally covered. **OoT's consume path does NOT re-dispatch `OnLoadGame`.** OoT's `OnLoadGame`-rebuilt tables (the ~20 rando gameplay hooks + the patched `gEntranceTable`) are safe only because the boot chain never clears them and the switch-out never tears them down. The two exceptions are OoT `initFunc` clears that run on the arrival title-demo `InitFile` and target module globals *outside* `gSaveContext` — those are the live bugs.

## OoT switch-arrival chain

`Title_Init` → `Title_Destroy` → `OoT_Sram_InitSram` → `Save_Init` (=`SaveManager::Init`); `Opening_Init` → `Opening_SetupTitleScreen` → `OoT_Sram_InitDebugSave` → `Save_InitFile(true)` → `initFuncs`; then `OoT_Play_Init` → `OoT_Play_ConsumeStartupEntrance` (restores `gSaveContext` bytes + `OoT_ConsumeSharedItems`; **no** `OnLoadGame` dispatch, **no** `LoadRandomizer`).

[...]

Issue #462, opened 2026-07-21
Converge the remaining 26 cross-game CVar rows from the classification inventory


spencerduncan:
Follow-up to #34 (PR #461), which landed the first 9 convergence rows: the audio-volume family and the three tunic-colour keys.

## What is left

`docs/enhancement-classification.md` §5.3 enumerates **26 further convergence rows** that #461 did not attempt. They are near-disjoint from what #461 landed — the inventory classified MM's `gEnhancements.*`/`gCheats.*` categories plus the 35-key collision set, while #461's audio family is in neither. The two sets intersect only on the three tunic keys.

- **Group A (12)** — identical leaf, MM adds a category prefix. Mechanical, highest confidence.
- **Group B (7)** — wording drift, same meaning. **2 of these are parked, see below.**
- **Group C (4)** — cutscene-skip family.
- **Group D (3)** — free-look, `gSettings.` vs `gEnhancements.Camera.`.

Direction is MM to OoT per ADR 0003 decision (3).

## Blocked rows — do NOT converge until answered

Four rows are parked on unanswered taxonomy questions. All four are already asserted as must-stay-distinct in `RSBS::kMustStayDistinct` (`src/common/cvar_shared_keys.h`), so `--test cvar-classification` fails if one is converged before its question is settled. **Resolving these is a prerequisite for the rows they belong to, not for the whole issue.**

1. `gCheats.InfiniteAmmo` vs `gCheats.InfiniteConsumables` — MM's "consumables" scope may be wider than OoT's "ammo".
2. `gCheats.NoRestrictItems` vs `gCheats.UnrestrictedItems` — same intent (drop item gating) but **different gate**: OoT ignores age restrictions, MM lets all forms use all items.
3. `gEnhancements.TimeSavers.` (OoT, capital S) vs `gEnhancements.Timesavers.` (MM, lowercase s) — they do not collide, but they are one keystroke apart in a case-sensitive store. Which spelling wins needs an explicit decision.
4. `gEnhancements.Saving.AutosaveInterval` — the Autosave **enable** flag is genuinely shared-intent and can converge, but OoT's interval is hardcoded to `THREE_MINUTES_IN_UNIX` with no CVar. The interval is MM-only and must NOT ride along with an `Autosave` rename.

## Also out of scope here

[...]

Issue #451, opened 2026-07-21
Four menu-index CVars are shared but index into two different menus (gates the 2ship_enh WHOLE_ARCHIVE flip)


spencerduncan:
Found while measuring cross-game CVar collisions for ADR 0003 (`docs/adr/0003-settings-namespace.md`).

## The defect

Four CVar keys are read by **both** games but their values name or index into **two different menus**:

- `gSettings.Menu.ActiveHeader`
- `gSettings.Menu.*SidebarSection` (the sidebar-section persistence keys)

Under ADR 0003 the project unifies on shared keys by default — but these are **class (P): shared key, divergent semantics**. A menu index is only meaningful relative to the menu it indexes. OoT's `SohMenu` and MM's `BenMenu` have different section/sidebar sets, so one game's persisted index selects an unrelated (or out-of-range) entry in the other.

This is the same hazard class as the #356 entrance-id leak and ADR 0002's untagged-`uint16_t` reasoning: an integer that is *almost* meaningful across a boundary.

## Why it is latent today

MM's menu is link-elided, so nothing on the MM side currently reads these keys.

## Why it is a sequencing gate

**It arms the moment `2ship_enh` gets `WHOLE_ARCHIVE`d** (the enhancement-surfacing chain, #427 item 2). Per ADR 0003, that flip is safe under the new policy *except* for these four keys.

So before the flip lands, either:
1. resolve the four keys (disambiguate per game — the ADR's class-(P) treatment, since these must NOT be merged), or
2. confirm in writing that MM's menu stays elided through the flip, and re-gate this issue on whenever it stops being.

## Related

- ADR 0003 (`72fd1898`) — classification, and the guard against "fixing" the 26 deliberate shared keys
- #34 — the convergence/migration work (8 MM→OoT renames); **this issue is the opposite treatment and must not be folded into that rename list**
- #427 item 2 — the `2ship_enh` migration + guard flip this gates
- #446 — the `Ship::Menu`/`MenuTypes` ODR landmine in the same menu area

spencerduncan:
**Likely discharged by ADR 0004 — re-gating rather than closing blind.**

[...]

PR #567, opened 2026-07-31
Re-apply MM's sound mode from the restored save at the arrival (#483)

### LINKED ISSUE #483: Arrival: MM sSoundMode derived from the bootstrap save, not re-applied at consume (low/self-healing — audit #482 row M5)
Spun out of the arrival-rehydration audit #482 (row M5) — a **weak/low-severity** instance of the class, filed rather than fixed because the symptom self-heals and the fix touches audio.

## What

On a cross-game switch INTO MM, the boot-init chain runs `MM_Sram_InitSram` → `Audio_SetFileSelectSettings` (`games/mm/src/code/z_sram_NES.c:1999`, reached from `ConsoleLogo_Destroy`, `z_title.c:253`), which writes the audio global `sSoundMode` from `gSaveContext.options.audioSetting`.

At that point in the boot chain the live options are the **vanilla bootstrap** save's (STEREO default), not the player's restored save. `sSoundMode` lives **outside `gSaveContext`**, and `MM_Play_ConsumeStartupEntrance` does not re-apply it after `Combo_ConsumeFrozenState` restores the real `options.audioSetting`. So immediately after an arrival, `sSoundMode` reflects the bootstrap setting, not the player's.

## Why it's low priority

The arrival scene's audio bring-up re-derives the sound mode, so the wrong value is transient and self-heals within a scene. A player who set mono/headset audio might get one moment of stereo on the arrival frame. This is the lowest-severity row in #482 and the only MM out-of-save *derived* write the consume path skips — every substantial MM derived table is rebuilt by the `OnSaveLoad` re-dispatch (#447).

## If fixed

Minimal inverse: re-apply `Audio_SetFileSelectSettings(gSaveContext.options.audioSetting)` (or the sound-mode equivalent) at `MM_Play_ConsumeStartupEntrance` after the restore, alongside the existing audio-neutralization there (`seqId`/`ambienceId` reset). Coordinate with the CVar/settings convergence work (#462) since MM audio settings are in flight there. Lock by driving the consume/audio-apply with a restored non-default `audioSetting` and asserting `sSoundMode` matches.

Audit: #482. Follow-up: #392.

spencerduncan:
Fixes #483. Lane D of the #564 one-game-semantics alignment wave — though #564 itself lists #483 under *Unaffected*: MM's `SaveOptions` is a per-install preference, not world identity. So this is the arrival-rehydration bug it always was (#482 row M5, the last open row), fixed on its own terms.

## The bug

`sSoundMode` (`games/mm/src/audio/code_8019AF00.c:224`) and the audio thread's copy live **outside `gSaveContext`**, so the frozen-save memcpy cannot carry them. On a cross-game arrival the boot chain derives them exactly once:

```
ConsoleLogo_Destroy            (z_title.c:253)
  -> MM_Sram_InitSram          (z_sram_NES.c:2180)
     -> Audio_SetFileSelectSettings(gSaveContext.options.audioSetting)
...
MM_Play_Init
  -> MM_Play_ConsumeStartupEntrance
     -> Combo_ConsumeFrozenState   <-- the player's options arrive HERE
```

The derivation runs **before** the restore, so it reads the vanilla bootstrap options (`SAVE_AUDIO_STEREO`), never the player's.

## #483's "self-healing" framing is wrong at source

The issue filed this as transient — "the arrival scene's audio bring-up re-derives the sound mode". It does not. `Audio_SetFileSelectSettings` is:

- the **only** writer of `sSoundMode` in the tree, and
- the only producer of `SEQCMD_SET_SOUND_MODE`, itself the only producer of `AUDIOCMD_GLOBAL_SET_SOUND_MODE` — so also the only path to the audio thread's copy.

Its other two call sites are the file-select Options submenu (`z_file_nameset_NES.c:779`) and the file-select save enumeration (`z_sram_NES.c:1944`). A cross-game arrival passes through neither. The bootstrap's mode was live for the **whole MM session**.

## Scope: contract now, audible later

Nothing in the combo currently moves `options.audioSetting` off the STEREO default — MM's Options submenu lives on the file select screen and an arrival fast-paths past it (`ConsoleLogo_Main` hands straight to `MM_TitleSetup_Init` when a startup entrance is pending). So today the re-apply re-asserts the value the boot chain already set. What lands is the contract: the arrival stops being the one place a live `options.audioSetting` is silently dropped, which is what the settings-convergence direction (#462, #451's one settings surface) needs, and #482's last open row closes either way.

## Why here and not on the #447 `OnSaveLoad` re-dispatch

The lane brief asked this explicitly. It is not the right home:

- That re-dispatch re-arms **save-shape-dependent** state — `COND_HOOK`s conditioned on `IS_RANDO`, read off `gSaveContext.save.shipSaveInfo.saveType`.
- No `OnSaveLoad` handler derives audio at all. There are exactly four registrations in MM: `Rando.cpp:41` (`OnSaveLoadHandler` → `MiscBehavior`/`ActorBehavior`/`CheckTracker`/`ClockShuffle::OnFileLoad` + `ShipInit::Init("IS_RANDO")`) and three in `SavingEnhancements.cpp` (`:255`, `:284`, `:333`). None touch the audio globals.
- `SaveOptions` is not save state: `SaveContext+0x3F40`, outside `Save`, outside the per-file blob the port serializes.
- Hanging a global preference off a file-load event would newly re-apply audio on every *other* `OnSaveLoad` dispatch (file select, map select, warp point) for no gain.

`MM_Play_ConsumeStartupEntrance` already owns exactly this responsibility for the arrival's other out-of-save audio globals (`seqId`, `ambienceId`, `forcedSeqId`). `sSoundMode` is the same shape and joins them, immediately after them.

## Range check

`Audio_SetFileSelectSettings`' switch has **no `default` case that assigns `soundMode`**, so an out-of-range `audioSetting` leaves `sSoundMode` standing but still runs `SEQCMD_SET_SOUND_MODE(soundMode)` on an uninitialised local — and `AudioSeq_ProcessSeqCmd` then indexes `sSoundModeList[5]` (`sequence.c:33`) with that payload. A restored blob, unlike the freshly-initialised options block `MM_Sram_InitSram` sees, is not guaranteed in range. Out of range we queue nothing and leave the boot chain's mode — exactly today's behavior on this path.

## Lock (ROM-free, `redship` tier)

Extends the existing `mm-startup-restore` row (`games/mm/2s2h/mm_resume_state_test.cpp`), which already drives the real `MM_Play_ConsumeStartupEntrance`:

| leg | arrange | assert |
|---|---|---|
| restored non-default | frozen save with `audioSetting = SAVE_AUDIO_HEADSET`; wipe; derive the bootstrap mode as `ConsoleLogo_Destroy` does (asserted to land on `SOUNDMODE_STEREO`, so nothing below passes vacuously) | `sSoundMode == SOUNDMODE_HEADSET` **and** the consume queued `SET_SOUND_MODE(SOUNDMODE_HEADSET)` for the audio thread |
| first entry | no frozen state, zeroed save | tracks the bootstrap mode (`SOUNDMODE_STEREO`) — re-derives, never invents. Non-vacuous: the previous leg left `HEADSET` live |
| out of range | frozen `audioSetting = 0x5A`, bootstrap `SAVE_AUDIO_MONO` | mode stays `SOUNDMODE_MONO` **and no `SET_SOUND_MODE` was queued** |

The third leg is asserted on the **command queue**, not on `sSoundMode`. That is deliberate: the `default` branch does not write `sSoundMode` either way, so an `sSoundMode`-only assertion there is vacuous — it would still pass with the range check deleted. The queue observation is the only one that fails.

## Verification (local, ROM-staged worktree)

Both tiers green on the final binary:

- `ctest -L "^redship$"` — **65/65 passed** (11.96 s)
- `ctest -L "^rando$"` — **15/15 passed** (27.41 s)
- MSVC/Ninja Release build clean; `clang-format` 14.0.6 `--dry-run -Werror` clean on both enforced files (`games/mm/src/code/z_play.c`, `games/mm/2s2h/mm_resume_state_test.cpp`).

**Both halves of the fix were falsified by rebuild-and-run, not asserted:**

| mutation | result |
|---|---|
| whole re-apply removed (`if (0)`) | `FAIL: restored options.audioSetting not re-applied to sSoundMode (#483)` at `mm_resume_state_test.cpp:239` |
| range guard removed (`if (1)`) | `FAIL: out-of-range restored audioSetting still queued a seq command (#483)` at `mm_resume_state_test.cpp:305` |

Note *which* assertion caught the second one: the queue-depth check, not `sSoundMode`. `sSoundMode` stayed `SOUNDMODE_MONO` in that run, exactly as predicted — which is why the `sSoundMode`-only version of that leg would have been a vacuous lock.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8784455459.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8784074850.zip)

### DIFF
--- a/games/mm/2s2h/mm_resume_state_test.cpp
+++ b/games/mm/2s2h/mm_resume_state_test.cpp
@@ -47,6 +47,18 @@ void Combo_ClearFrozenState(const char* gameId);
 void Combo_SetStartupEntrance(uint16_t entrance);
 bool Combo_HasStartupEntrance(void);
 void Combo_ClearStartupEntrance(void);
+// games/mm/src/audio/code_8019AF00.c — the derived sound-mode global the
+// arrival must re-apply from the restored save (#483). Not declared in any
+// header; Audio_SetFileSelectSettings (declared in sequence.h, reached through
+// global.h) is its only writer.
+extern s8 sSoundMode;
+// games/mm/src/audio/sequence.c — write cursor into MM_sAudioSeqCmds (declared
+// in sequence.h). Nothing drains the queue in a headless run, so the delta
+// across a call is exactly what that call handed the audio thread. Needed
+// because sSoundMode alone cannot see the second half of the #483 write:
+// Audio_SetFileSelectSettings' default branch leaves sSoundMode untouched but
+// still queues SEQCMD_SET_SOUND_MODE with an unassigned soundMode.
+extern u8 sSeqCmdWritePos;
 }
 
 extern "C" u8* MM_gSystemHeap;
@@ -60,6 +72,45 @@ namespace {
             return 1;                                                     \
         }                                                                 \
     } while (0)
+
+/**
+ * How many seq commands were queued since write cursor `from`.
+ *
+ * This is the half of the sound-mode write sSoundMode cannot report.
+ * Audio_SetFileSelectSettings has no `default` case that assigns soundMode, so
+ * an out-of-range audioSetting leaves sSoundMode standing — indistinguishable
+ * from not calling it at all — while still queueing a command whose payload the
+ * audio thread uses to index sSoundModeList[5]. A DEPTH check is what catches
+ * that, and it is deliberately payload-blind: the queued word is
+ * `... | (soundMode)` on an `s8`, so an unassigned NEGATIVE soundMode
+ * sign-extends over the opcode nibble and an opcode-matched probe would miss
+ * the very command it exists to catch.
+ */
+u8 SeqCmdsQueuedSince(u8 from) {
+    return (u8)(sSeqCmdWritePos - from);
+}
+
+/**
+ * The SoundMode payload of the last SEQCMD_SET_SOUND_MODE queued since write
+ * cursor `from`, or -1 if none was queued. Sound only for in-range values (see
+ * SeqCmdsQueuedSince) — used to check WHICH mode a good call asked for.
+ */
+int LastQueuedSoundModeSince(u8 from) {
+    // Same op/sub-op mask-and-value shape AudioSeq_IsSeqCmdNotQueued uses:
+    // op in bits 31..28, sub-op in bits 11..8, payload in bits 7..0.
+    const u32 kMask = 0xF0000F00;
+    const u32 kSetSoundMode = ((u32)SEQCMD_OP_GLOBAL_CMD << 28) | ((u32)SEQCMD_SUB_OP_GLOBAL_SET_SOUND_MODE << 8);
+    int found = -1;
+
+    for (u8 i = from; i != sSeqCmdWritePos; i++) {
+        u32 cmd = MM_sAudioSeqCmds[i];
+
+        if ((cmd & kMask) == kSetSoundMode) {
+            found = (int)(cmd & 0xFF);
+        }
+    }
+    return found;
+}
 
 } // namespace
 
@@ -113,8 +164,11 @@ extern "C" int MM_ResumeArena_RunHeadless(void) {
 /**
  * mm-startup-restore: MM_Play_ConsumeStartupEntrance must (a) restore the
  * frozen save over a boot-chain wipe, (b) spawn at the startup entrance with
- * cutscene/game-mode state reset, (c) clear the startup slot, and (d) leave
- * a wiped save alone when there is no frozen state (first entry).
+ * cutscene/game-mode state reset, (c) clear the startup slot, (d) leave
+ * a wiped save alone when there is no frozen state (first entry), and (e)
+ * re-apply the sound mode DERIVED from the restored save's
+ * options.audioSetting, which lives outside gSaveContext and so cannot ride
+ * the restore memcpy (#483, audit #482 row M5).
  */
 extern "C" int MM_StartupRestore_RunHeadless(void) {
     printf("[TEST] mm-startup-restore: startup consumption restores the frozen save post-wipe\n");
@@ -126,17 +180,29 @@ extern "C" int MM_StartupRestore_RunHeadless(void) {
     memset(&gSaveContext, kPattern, sizeof(gSaveContext));
     gSaveContext.save.day = 3;
     gSaveContext.save.time = 0x4321;
+    // A NON-DEFAULT sound preference in the frozen save. The 0x5A fill would
+    // otherwise leave audioSetting out of range, which is a separate case
+    // (locked at the end of this function).
+    gSaveContext.options.audioSetting = SAVE_AUDIO_HEADSET;
     ((uint8_t*)&gSaveContext)[sizeof(gSaveContext) - 1] = 0x77;
     Combo_FreezeState("mm", kArrival, &gSaveContext, sizeof(gSaveContext));
 
     // The boot chain's wipe (Setup_InitImpl -> MM_SaveContext_Init).
     memset(&gSaveContext, 0, sizeof(gSaveContext));
 
+    // ...and the boot chain's derived sound mode, taken from the VANILLA
+    // BOOTSTRAP options the wipe just left behind (ConsoleLogo_Destroy ->
+    // MM_Sram_InitSram -> Audio_SetFileSelectSettings, z_sram_NES.c). This
+    // runs BEFORE the restore below — which is the whole bug.
+    Audio_SetFileSelectSettings(gSaveContext.options.audioSetting);
+    RESUME_ASSERT(sSoundMode == SOUNDMODE_STEREO, "bootstrap-derived sound mode not the STEREO default");
+
     // The switch path's pending startup entrance (wildcard tag — visible to
     // MM, same visibility rule the game-scoped setter grants).
     Combo_SetStartupEntrance(kArrival);
 
     // Act: the consumption point in MM_Play_Init.
+    const u8 seqCmdPosBeforeArrival = sSeqCmdWritePos;
     MM_Play_ConsumeStartupEntrance();
 
     // Assert: frozen save is back...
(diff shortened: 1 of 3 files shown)


spencerduncan:
## Adversarial review — independently verified, no changes required

Reviewed at `406093cc` against `origin/main` (`342c515d`). Everything load-bearing was re-derived from the tree rather than taken from the implementer's report.

### Declarations re-checked against the real sources

| claim | verified |
|---|---|
| `sSoundMode` is a plain (non-static) `s8` global | `games/mm/src/audio/code_8019AF00.c:224` — matches the test's `extern s8 sSoundMode;` |
| `Audio_SetFileSelectSettings` is its only writer | 6 references; only the four `case` arms at `:6105/:6110/:6115/:6120` assign |
| exactly three call sites, two file-select-only | `z_sram_NES.c:2183` (`MM_Sram_InitSram`), `z_sram_NES.c:1944` (file-select enumeration), `z_file_nameset_NES.c:779` (Options submenu) |
| derivation runs before the consume | `ConsoleLogo_Destroy` (`z_title.c:250-253`) → `MM_Sram_InitSram`; the consume is in `MM_Play_Init` |
| `default:` assigns no `soundMode`, and `SEQCMD_SET_SOUND_MODE(soundMode)` runs unconditionally after the switch | `code_8019AF00.c:6099-6128` |
| the payload indexes a 5-element array | `sSoundModeList` (`sequence.c:33`), consumed at `sequence.c:403` |
| the range check has the right bound and the right signedness | `options.audioSetting` is `u8` (`z64save.h:299`), `AudioOption` is `0..3` (`z64save.h:17-21`), so `<= SAVE_AUDIO_SURROUND` rejects both `0x5A` and `0xFF` |
| `sSeqCmdWritePos` is a non-static `u8` and `MM_sAudioSeqCmds` is `[0x100]`, so the `u8` cursor arithmetic is exact | `sequence.c:30`, `sequence.h:413` |
| no new cross-game ODR surface | OoT's twins are *differently named* (`sSeqCmdWrPos`, `OoT_sAudioSeqCmds`) and OoT has no `sSoundMode` at all |
| #564 really does classify #483 as out of scope for the one-game ruling | #564 body, "Unaffected (checked, no change)": "#483 (per-install audio preference, not identity)" |

### The one-game "both legs" check

The obvious reviewer objection to a one-sided fix is that OoT's return leg has the same shape — `OoT_Sram_InitSram` (`z_sram.c:317-321`) calls `func_800F6700(gSaveContext.audioSetting)`, which writes the out-of-save `D_80130604`, and it is dispatched from `Title_Destroy` (`z_title.c:172-176`) on every return through the title chain.

It is **not** the same bug, and the asymmetry is real rather than an oversight. On OoT's return leg the restore happens *before* the derivation: `OoT_Game_Resume` runs `Context_RestoreState` (`GameExports_SingleExe.cpp:921-924`), and the clobber that motivates OoT's own consume — `Opening_Init` re-authoring the ADULT debug save — happens *after* `Title_Destroy`. `Save_Init()` does not touch `gSaveContext.options`. So OoT's sound mode is derived from the restored `audioSetting`. MM's inverts that order (`Setup_InitImpl` → `MM_SaveContext_Init` memsets *after* `MM_Game_Resume`), which is exactly why only MM needs the re-apply.

### On leaving the corrupt byte in the live save

The guard skips the call but does not normalise an out-of-range `options.audioSetting`, so the bad byte stays in the restored save. Chased this as a possible latent hazard and it does not reach anything: the only other in-combo reader is `z_sram_NES.c:1944`, and that path `memcpy`s the whole `SaveOptions` block out of the flash buffer (and resets it wholesale when `optionId != 0xA51D`) *before* calling `Audio_SetFileSelectSettings`, so the live value is overwritten first. Vanilla's own integrity gate is that `optionId` magic, not a range check — mirroring it here would mean re-authoring restored per-install preferences, which is more than this bug needs. The narrow guard is adequate.

### Counterfactuals re-run locally by the reviewer

Not taken on trust — both mutations were rebuilt and run on this workstation (throttled `-j4`, `start /low`, one build at a time), in the lane's ROM-staged worktree, and the tree was restored and rebuilt afterwards.

| mutation | result |
|---|---|
| `if (0 && ...)` — whole re-apply removed | `[TEST] FAIL: restored options.audioSetting not re-applied to sSoundMode (#483) (mm_resume_state_test.cpp:239)` |
| `if (1)` — range guard removed | `[TEST] FAIL: out-of-range restored audioSetting still queued a seq command (#483) (mm_resume_state_test.cpp:305)` |

The second one failed on the **queue-depth** assertion at `:305` while the `sSoundMode == SOUNDMODE_MONO` assertion at `:303` passed — independently confirming the anti-vacuity claim in the PR body: an `sSoundMode`-only version of that leg could not have failed.

Restored implementation, same worktree, final binary:

- `ctest -L "^redship$"` → **100%, 65/65** (15.59 s), `MMStartupRestore` passing
- `ctest -L "^rando$"` → **100%, 15/15** (24.01 s)

clang-format 14.0.6 `--dry-run -Werror` clean on both enforced paths.

### Base freshness

The branch is based on `ec510ec6`; `main` has since advanced to `342c515d` (#566 — libultraship archive-handle mutex + the `ZipContention` row). `git merge-tree` against current `main` is conflict-free, and #566 touches only the archive/resource layer and a disjoint region of `test_runner.cpp` — no interaction with the MM audio path. Not rebased, deliberately: rebasing would drag the in-flight submodule pointer into this branch for no review value.

**Verdict: merge-ready as-is.** No defects found, no changes pushed.

Issue #498, opened 2026-07-22
Combo-level settings: direction, pool size, and item classes have no storage, no gate, and no world-identity term


spencerduncan:
Every cross-game randomization parameter is currently a compile-time constant: the direction is hardcoded OoT→MM, the "pool" is a 4-element `static const` array, and the item class is "pinned OoT progression". There is no CVar, no `ComboContext` field, and no menu control that governs any of it, so a player cannot choose what crosses and two peers cannot disagree about it. This issue covers making those three things *settings* — which is mostly a format/ordering decision, and only secondarily code.

## What is actually hardcoded (KNOWN)

- The pool is four literal entries with a `sizeof`-derived count and a `static_assert` against the placement carve: `games/oot/soh/Enhancements/randomizer/ForeignItemsSingleExe.cpp:58-67`. That TU's header comment is the constraint any generalisation must respect — it is the one translation unit where real `RG_*` enumerators may appear (ADR 0002).
- Placement policy is hardcoded too: candidates are non-`RCTYPE_SHOP`/`RCTYPE_TINGLE_SHOP` checks whose `RANDO_SAVE_CHECKS[id].shuffled` bit is set and whose held item is `RITYPE_JUNK`, selected by a local xorshift seeded from `sharedRandoSeed + sharedRandoSettingsHash + rando.finalSeed + ":foreign-v1"` — `games/mm/2s2h/Rando/Foreign.cpp:117-152`.
- Direction is structural, not a flag. The placement table's key is literally `uint16_t mmCheckId` and its 6-byte layout and member offsets are `static_assert`ed as .redsave format (`src/common/context.h:233-242`), so it cannot gain a `hostGame` byte in place.
- The gate that a placement pass reads, `Combo_ForeignPairingActive()`, is `gComboCtx.sourceIsRando && sharedRandoSettingsHash != 0` (`src/common/foreign_items.c:18-20`), and those fields are stamped at the *end* of OoT's `Playthrough_Init` — `games/oot/soh/Enhancements/randomizer/3drando/playthrough.cpp:116-118`, after `Fill()` at `:81` and after the spoiler write at `:91`.
- The world identity carries OoT's settings only: `sharedRandoSettingsHash = SohUtils::Hash(settingsStr)` computed at `playthrough.cpp:71`. `ComboContext` has no field for combo settings and no field for MM's option profile (`src/common/context.h:330-364`).
[...]

PR #570, opened 2026-07-31
Freeze the MM profile into the pairing identity at creation; refuse divergent arrivals through the #533 surface (#498, #564)

spencerduncan:
Implements #564 phase 2 step 9 — the profile freeze + arrival compare-and-refuse that the one-game ruling made #498 decision 1's settled answer. Unstacked by the merged #533 REFUSED machinery (#568), which this change plugs its fourth refusal producer into.

## What changes

**1. The creation event stamps the MM profile identity (V1 minimal form, V3, V8).**
`Playthrough_Init` now freezes the MM half's option profile into the pairing identity at the same publish point that stamps `sourceIsRando` / `sharedRandoSeed` / `sharedRandoSettingsHash`: `gComboCtx.mmProfileDigest = MM_Rando_ComputeProfileStamp()`. The stamp is computed by the SAME `ResolveProfileValues` + `ProfileIdentityString` pair MM's arrival re-runs (one shared builder in `Foreign.cpp`), so creation and arrival cannot drift. The resolved `RO_LOGIC` pin value is what gets stamped — nothing after creation depends on CVar existence except through this one resolution (V3).

**2. The digest widens to the generator's full input set (V4).**
`ProfileIdentityString` folds the 47 resolved option values **plus** `gRando.ExcludedChecks` (the CVar `GeneratePools` consumes) **plus** the StartingItems config block (via the same `GetStartingItemsFromConfig` accessor generation calls, now null-guarded for the display-free harness so the identity term and the generation input stay one computation). `MMOptionsString` — the **seed-derivation** input `MixPairedFinalSeed` hashes — is deliberately untouched: widening it would re-derive every shipped pair's MM world.

**3. KEEP carries the stamp through file-create invalidation (V9).**
`Context_InvalidateSessionState` snapshots/restores `mmProfileDigest` with the seed stamp, per the "the rule, not the list" contract #565 wrote into `ComboSeedStampPolicy`. DROP paths clear it, so a dead pair's digest can't freeze the pane for a world that no longer exists.

**4. Post-creation writes are rejected; the pane renders frozen (V5).**
New src/common predicate `Combo_MMProfileFrozen()` (`gComboCtx.mmProfileDigest != 0` — a src/common fact, never a gSaveContext read, per ADR 0008 rule 5). Both option writers (`Combo_MMOptionSetValue`, `Combo_MMOptionClear`) reject while frozen; the pane draws ADR 0004 §6's fourth presentation state (frozen banner, all rows disabled, Reset disabled with inline reason) and the "set them before you cross" copy that taught the retired renegotiation window is gone.

**5. MM arrival compares and refuses (V8: the digest acquires comparators).**
`MM_Rando_PairOnCrossGameArrival` re-resolves the identity before dispatching generation. Mismatch => REFUSE through the #533 machinery: no `OnSaveInit` dispatch (never author a divergent world), stamp left untouched (never self-heal), the active slot latched + surfaced via new `RSBS_REFUSE_IDENTITY` (`RefuseSlotIdentity` latches **without quarantining** — the on-disk .redsave is healthy and is exactly what the latch protects from the divergent session's captures), plus an immediate toast on the shared overlay. `ResolvePairedProfile` throws on the same mismatch as the class-level backstop for the non-arrival generation routes (MM file-select escapes, dev warps). A `mmProfileDigest == 0` pair is a LEGACY pre-freeze pair and keeps the old freeze-at-first-crossing behavior — the one transitional writer besides creation.

**6. The remaining CVar families are CLASSIFIED, not deferred (V4's tail).**
`gRando.Traps.*` is **runtime flavour, not world identity**: its only consumer `RollTrapType()` is called from actor behaviour when a trap is collected (`ActorBehavior/EnGirlA.cpp:73`, `Traps.cpp:133`), drawing from the live `Ship_Random` stream at pickup — never inside a fill. The trap knobs that *do* shape the world (`RO_SHUFFLE_TRAPS`, `RO_TRAP_AMOUNT`, read at `GeneratePools.cpp:281`) are ordinary option rows and are already folded. `gRando.SpoilerFile` / `SpoilerFileIndex` / `InputSeed` are excluded and folding them would be actively harmful: the paired path ignores all three by construction (`OnFileCreate.cpp:83` forces generation over a stale spoiler index when paired; `:101` replaces the input seed with `PairedInputSeedString`), and `OnFileCreate.cpp:289` *writes* `gRando.SpoilerFile` after every successful generation — a pair that folded it would diverge its own identity the instant it generated. Recorded in `ProfileIdentityString`, because ADR 0004 §6's scope note defaults an unclassified key to identity: silence is not neutral.

## Locks (all falsifiable, all counterfactuals MEASURED not asserted)

| Lock | Where | Counterfactual that goes red |
|---|---|---|
| Divergent switch-entry arrival **surfaces** the refusal: slot REFUSED w/ `RSBS_REFUSE_IDENTITY`, write-latched, stamp intact | `mm-pair-switch-entry` phase 4 (rando tier) | Revert the compare in `MM_Rando_PairOnCrossGameArrival` **alone** -> FAIL(24)/FAIL(26) (measured 2026-07-31). Note: **not** FAIL(22) — the backstop below still stops the world from being authored, so this layer buys the *surface*, and reverting it degrades the refusal into #564 V7's silent vanilla revert |
| Divergent arrival authors **no world** | `mm-pair-switch-entry` phase 4 | Revert **both** compares (arrival gate + `ResolvePairedProfile`) -> FAIL(22) + `mm-paired-profile` FAIL(66) (measured 2026-07-31): the divergent leg generates a world under the mid-session-edited CVars |
| Stamped-mismatch resolution throws and never self-heals; creation-side stamp == arrival-side digest under identical inputs | `mm-paired-profile` (redship tier) | Revert the compare/stamp split in `ResolvePairedProfile` -> FAIL(66)/FAIL(67) |
| Excluded-checks edit alone moves the digest | `mm-paired-profile` | Narrow the identity string back to the option table (the pre-widening digest) -> FAIL(68) |
| Frozen writers reject; unfrozen writers work | `combo-mm-options-window` | Remove the `Combo_MMProfileFrozen()` gate from either writer |
| KEEP preserves the stamp across file-create invalidation; every DROP clears it | `session-invalidation` | Remove the `mmProfileDigest` snapshot/restore from the KEEP set |
| Identity refusal latches + surfaces without touching or quarantining the healthy slot file; erase releases it | `save-write-latch` | Remove the latch (or add a quarantine) in `RefuseSlotIdentity` |

## Verification

Independently re-verified at head `dbcaf843` on a fresh local build (Ninja/MSVC Release, submodules initialized, ROM-staged archives, worktree isolated from every other checkout):

- `ctest -L "^redship$"` — **100% tests passed out of 70**
- `ctest -L "^rando$"` — **100% tests passed out of 15** (including `RandoDeterminism`, `SeedDeterminism`, `MMRandoGen`, `ForeignPlacementOoT`, `MMPairSwitchEntry`)
- clang-format 14.0.6 clean on every enforced path this PR touches (`GameExports_SingleExe.cpp`, `Foreign.cpp`, `Foreign.h`).
- Both counterfactuals executed, not asserted (results in the table above). No submodule pointer and no generated stamp (`build.c`, `properties.h`) is in any commit.

Determinism check performed as part of review: `Ship_Hash` is pure FNV-1a and consumes no RNG, and nothing on the OoT reverse-placement path reads `mmProfileDigest`, so inserting the stamp before `OoT_PlaceForeignItems` cannot shift any existing pair's placement stream — confirmed by `SeedDeterminism`/`RandoDeterminism` staying green.

## Known consequences, stated rather than discovered later

- **The stamp lands at GENERATION, not at file-create.** That is what #564 step 1 requires (the freeze must precede OoT's `Fill()`, V24), and it mirrors OoT's own snapshot-at-generation precedent — but it means a player who generates a seed and only *then* wants different MM options is frozen with no file in existence yet, and re-generating re-stamps the same frozen profile. The only unfreezing events are `Context_InvalidateSessionState`'s DROP paths: returning to the title screen, starting a vanilla file, or loading a slot whose `.redsave` is unfrozen. The frozen banner now names that escape explicitly instead of saying "create a new paired world", which from the frozen state is advice the player cannot act on.
- **State 4 ships as the gate and the presentation, not yet as the frozen VALUES.** ADR 0004 §6 also requires a frozen key's value be shown *from the save*; the pane still renders live CVar reads. That is unimplementable under the interim hybrid, which persists only the digest (#564's carve budget shows a 47-option Tier-1 record does not fit). In-app the two agree — the writers are the only in-app writers and they reject — but an out-of-band write (hand-edited config, libultraship console `cvar_set`) makes the pane display a value the world was not built from, caught only at the next arrival. Named as residue in ADR 0004 §4.1a and owed to #564 step 11's move to delivery option 1.
- **#564's acknowledged hybrid cost stands:** a pair created-but-never-crossed whose inputs later change refuses its MM half until erased. Surfaced loudly, latched, recoverable via erase.
- **`gRando.ExcludedChecks` is folded as the raw CVar string**, so any mid-pair write to it is divergence by definition. In single-exe this is safe by construction: its only writer, MM's `Rando/Menu.cpp`, is link-elided (`games/mm/CMakeLists.txt:141,:146`).

## Scope notes

Part of #498 / #564 phase 1 (in #564's numbering: phase 2 step 9, the freeze's minimal form). **Closes nothing yet** — #498 retains real scope beyond this: the combo-level settings tier itself (`comboSettingsHash`, tier-4 keys, decision 3's display-name collision), and #564's step 11 (merged generation at creation, at which point arrival loses its generation dispatch entirely and this compare becomes hydrate-or-refuse). The one-spoiler identity (V23) remains out of scope.

Avoided per lane constraints: the #532/#543 `TitleSetup` mechanism (untouched — the gate lives in the #439 `MM_Rando_PairOnCrossGameArrival` dispatch) and the libultraship submodule pointer.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8800498102.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8800164318.zip)

### DIFF
--- a/docs/adr/0004-menu-information-architecture.md
+++ b/docs/adr/0004-menu-information-architecture.md
@@ -256,6 +256,21 @@ stamps `gComboCtx.mmProfileDigest`, the two `src/common` option writers reject a
 renders the frozen banner with every row disabled. The predicate is `Combo_MMProfileFrozen()`,
 the `src/common` fact §6 prescribes.
 
+**Residue, named rather than implied: state 4 is implemented as the gate and the presentation, not
+yet as the frozen VALUES.** §6 also requires that where a frozen key's value is shown, it be the
+value *from the save* rather than the CVar. This pane still renders live CVar reads. That is
+correct-but-fragile today — the two `src/common` writers are the only in-app writers and they
+reject while frozen, so the two agree — and it is genuinely unimplementable under the interim
+hybrid: the hybrid persists only the digest, so there is no frozen profile in `src/common`-owned
+storage to publish (§6's own "publishing the frozen profile into a `src/common` view is part of the
+freeze work" is the ask, and #564's carve budget shows a 47-option Tier-1 record does not fit).
+An out-of-band write the writers never see — a hand-edited config, libultraship's console
+`cvar_set` — therefore makes the pane display a value the world was not built from while labelled
+"frozen at creation". It is caught (the next arrival refuses on the digest) but it is displayed
+wrongly until then. Closing this is part of #564 step 11's move to delivery option 1, where the
+frozen profile lives in MM's Tier-3 `RANDO_SAVE_OPTIONS` and reaches the pane through the same
+`src/common` accessor the digest already uses.
+
 > **Amended 2026-07-30 (#564): the host conclusion survives and gets stronger; the deadline is
 > restated.** "Reachable while OoT is the running game, before the switch" was the right host
 > argument off the wrong deadline. Under one-game semantics the MM option profile freezes at the
--- a/docs/adr/0004-menu-information-architecture.md
+++ b/docs/adr/0004-menu-information-architecture.md
@@ -249,11 +249,12 @@ arming condition is untouched (mechanized as the MM-side reader allowlist in
 
 **Consequence accepted: §4.2's shared-intent marker does not apply to this pane.** All 47 options
 are tier-3 (O) MM-only keys; none is in `kSharedIntentKeys`, so there is no "applies to both
-games" claim to mark. §6's presentation states DO apply: live, editable-but-not-active
-("Majora's Mask — suspended"), and disabled-by-capability with a reason are implemented; §6's
-fourth state (frozen-at-creation, read-only) applies to this pane too and is **not** implemented
-— it is owed with the freeze work (#564 V5), and until it exists the pane is a live CVar editor
-at all times, including after the world it describes has been created.
+games" claim to mark. §6's presentation states DO apply, all four: live, editable-but-not-active
+("Majora's Mask — suspended"), disabled-by-capability with a reason, and — implemented with the
+#564-V5 freeze work (#498 phase 2 step 9) — frozen-at-creation, read-only: once a creation event
+stamps `gComboCtx.mmProfileDigest`, the two `src/common` option writers reject and the pane
+renders the frozen banner with every row disabled. The predicate is `Combo_MMProfileFrozen()`,
+the `src/common` fact §6 prescribes.
 
 > **Amended 2026-07-30 (#564): the host conclusion survives and gets stronger; the deadline is
 > restated.** "Reachable while OoT is the running game, before the switch" was the right host
--- a/games/mm/2s2h/GameExports_SingleExe.cpp
+++ b/games/mm/2s2h/GameExports_SingleExe.cpp
@@ -40,6 +40,12 @@
 // Paired-world keying + placement-table accessors (#439 switch-entry
 // activation logs the placement count at the pairing decision point).
 #include "foreign_items.h"
+// MM_Rando_ComputeProfileStamp — the arrival's identity re-resolution
+// (#498/#564: compare against the creation-frozen mmProfileDigest).
+#include "combo_mm_options_view.h"
+// OoT_Notification_Emit — the shared toast overlay (#427 bridge): the arrival
+// refusal must be player-visible in-game, not only in the file panel.
+#include "notification_bridge.h"
 #include "entrance.h"
 #include <ship/resource/ResourceManager.h>
 #include <ship/resource/ResourceLoader.h>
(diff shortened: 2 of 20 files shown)


spencerduncan:
Local verification on the Windows workstation (Ninja/MSVC Release, full ROM-staged worktree build at this branch head):

- `ctest -L ^redship$`: **70/70 passed** (12.2s) — includes the updated SessionInvalidation, SaveRefusedQuarantine/SaveWriteLatch/SaveArmOnCreate/SaveRefusedMeta, ComboMMOptionsWindow, MMRandoOptions, and MMPairedProfile rows.
- `ctest -L ^rando$`: **15/15 passed** (24.3s) — includes RandoDeterminism + SeedDeterminism (OoT generation now stamps the profile digest in-process and the paired MM generation's compare passes), MMRandoGen, ForeignPlacementOoT, and MMPairSwitchEntry with the new phase 4.

MMPairSwitchEntry phase-4 markers from the run, proving the lock is non-vacuous in both directions:

```
[MM-PAIR-SWITCH] divergent arrival refused: no world, stamp intact, slot latched + surfaced
[MM-PAIR-SWITCH] matching arrival generated normally under the frozen profile
```

clang-format 14 clean on the enforced paths (`GameExports_SingleExe.cpp`, `Foreign.cpp`, `Foreign.h`).

spencerduncan:
## Adversarial review — verified at `dbcaf843`

Re-reviewed the full diff against the real declarations, re-ran both counterfactuals myself on a fresh isolated build, and pushed four corrections. The gate's behavior is unchanged; what changed is that the claims about it are now true.

### What held up

- **The gate sits in the right place.** It is after both skip legs (`hadFrozenState`, `alreadyRando`), so it only runs on the arrival that would *generate*. A second crossing over an existing MM half takes the skip and never compares — no false refusal on subsequent crossings, which was the failure mode I went looking for first.
- **No generation path writes an identity term.** Grepped every `CVarSet*` under `2s2h/Rando/`: the only writes on the paired path are `gRando.SpoilerFile` / `SpoilerFileIndex`, neither folded. Had spoiler CVars been folded (the "obvious" widening), every pair would have diverged its own identity at `OnFileCreate.cpp:289`.
- **No existing pair's world re-derives.** `Ship_Hash` is pure FNV-1a with no RNG consumption, `MMOptionsString` is untouched, and nothing on the reverse-placement path reads `mmProfileDigest` — so inserting the stamp before `OoT_PlaceForeignItems` cannot shift a placement stream. `SeedDeterminism` and `RandoDeterminism` green confirm it.
- **`values[RO_MAX]`** matches `u32 randoSaveOptions[RO_MAX]` (`z64save.h:387`); `MM_Rando_ComputeProfileStamp` is declared inside `combo_mm_options_view.h`'s `extern "C"` block, so the OoT-side call links to the MM-side definition by C linkage rather than coincidence.
- **`Combo_MMProfileFrozen()` is a `src/common` fact**, not a `gSaveContext` read — ADR 0008 rule 5 satisfied as #564 V5 requires.

### What did not, and is now fixed

**1. The headline counterfactual was false. I measured it.** The claim was "revert the compare in `MM_Rando_PairOnCrossGameArrival` and the divergent arrival generates a different paired world → FAIL(22)". Reverting it alone:

```
[MM] pairing: armed on switch-entry (...) — dispatching OnSaveInit
[MM] pairing: FAILED — generation reverted the save to vanilla
[MM-PAIR-SWITCH] FAIL(24): the refused arrival did not surface REFUSED on the active slot (state=0)
```

FAIL(22) stayed **green** — `ResolvePairedProfile`'s backstop still threw and `OnFileCreate`'s catch still reverted to vanilla. Reverting **both** compares is what produces the advertised failure:

```
[MM-PAIR-SWITCH] FAIL(22): a DIVERGENT arrival generated a world — a mid-session option edit changed the paired world instead of being refused
[MM-RANDO-OPTIONS] FAIL(66): a stamped pair accepted a DIVERGENT profile resolution
```

The lock is real and *stronger* than described — two independently falsifiable layers — but the description was wrong about which assertion measures what. The arrival gate's job is the **surface**: revert it and the refusal degrades into #564 V7's silent vanilla revert, with no REFUSED state, no reason, and no write latch on the healthy pair. The test comment now states the measured result.

**2. #564 step 9's "classify trap types (V4)" was silently dropped.** ADR 0004 §6's scope note says an unclassified key *defaults to identity*, so leaving it out was a decision, not an omission. Classified with evidence in `ProfileIdentityString`: `gRando.Traps.*` is runtime flavour (`RollTrapType()` runs from actor behaviour at pickup off the live RNG, never inside a fill), while the world-shaping trap knobs are already-folded option rows. Same for the spoiler/InputSeed family, with the reason folding them would be harmful.

**3. ADR 0004 §4.1a overclaimed.** It said all four presentation states are implemented. §6 also requires a frozen key's value be shown *from the save*; the pane still renders live CVar reads, and cannot do otherwise under the hybrid (only the digest persists). In-app the writers keep the two in agreement, but an out-of-band write — hand-edited config, libultraship console `cvar_set` — makes the pane display a value the world was not built from while labelled "frozen at creation". Named as residue owed to step 11 rather than left as a V12-class overclaim.

**4. The frozen banner gave advice the player cannot follow.** "To play different MM options, create a new paired world" — but the stamp lands at *generation*, so a player who generates and then wants different options is frozen with no file in existence, and re-generating re-stamps the same profile. The only unfreezing events are `Context_InvalidateSessionState`'s DROP paths. The copy now names the title screen.

### Verification at `dbcaf843`

Fresh local Windows build in an isolated worktree (Ninja/MSVC Release, submodules initialized, ROM-staged archives):

- `ctest -L "^redship$"` → **100% tests passed out of 70**
- `ctest -L "^rando$"` → **100% tests passed out of 15**
- clang-format 14.0.6 clean on `GameExports_SingleExe.cpp`, `Foreign.cpp`, `Foreign.h`
- No submodule pointer, no generated stamp (`build.c`, `properties.h`) in any commit

Not merged; no issue closed.

Issue #500, opened 2026-07-22
[Tracker] Phase 3.2 — cross-game logic and beatability (Lane D promoted)


spencerduncan:
Lane D — cross-game logic — was deliberately deferred out of 3.0 (`docs/phase3-roadmap.md:195-201`: "not placing MM's Bow behind an OoT check that requires MM's Bow is the hardest problem in the phase"), with the spoiler log carrying the burden logic would otherwise carry. This tracker promotes it. Today the combo has **zero** cross-game logic scaffolding — not a stub, not a header, not a TODO hook — and the one thing that sounds like it already exists (`RandoHintCrossGame`) is not it.

The roadmap files this under "LATER — Phase 3.1+" as "Cross-game logic + beatability check (Lane D promoted)" (`docs/phase3-roadmap.md:207`); this tracker scopes it as 3.2 so that 3.1 can stay on the increments already in flight. Nothing here is committed until the individual issues are filed off this tracker.

## KNOWN (verified at origin/main, `8352cd4d`)

**Two complete, independent, single-game engines that never see each other's state.**

- OoT: static region graph `std::array<Region, RR_MAX> areaTable` with capture-less `ConditionFn` predicates and 4-state (childDay/childNight/adultDay/adultNight) reachability — `games/oot/soh/Enhancements/randomizer/location_access.h:22, :57-60, :122-149, :253`, populated by `RegionTable_Init()` at `location_access.cpp:958`.
- OoT **has** a beatability routine: `bool CheckBeatable(RandomizerGet)` crawls from the region pool with `stopOnBeatable=true` (`games/oot/soh/Enhancements/randomizer/3drando/fill.cpp:588-604`), wrapped per-item by `IsBeatableWithout()` (`fill.cpp:301-307`).
- OoT logic is **headless and save-independent**: `Logic::NewSaveContext()` allocates its own `new SaveContext()` (`games/oot/soh/Enhancements/randomizer/logic.cpp:2314`), called from `Logic::Reset(true)` at the top of every generation (`logic.cpp:2647-2650`).
- MM: runtime-built region graph `std::map<RandoRegionId, RandoRegion> Regions` with `std::function` predicates, populated by per-file `RegisterShipInitFunc` registrars (root node `RR_MAX` at `games/mm/2s2h/Rando/Logic/Logic.cpp:139-158`; traversal `FindReachableRegions` at `Logic.cpp:97-137`; 45-slice time model at `Logic.h:15-72`).
[...]

Issue #488, opened 2026-07-22
Foreign-host selection trusts fill-time `.shuffled` alone, so a pinned OoT item can land on a player-excluded check or an unresolved item slot (Lane C1 hardening, follow-up to #392)


spencerduncan:
`Rando::Foreign::PlaceForeignItems` picks the MM checks that host OoT's pinned foreign items using a predicate that says nothing about whether the host can ever hand its item over: it requires only `.shuffled` (a fill-time bookkeeping bit) plus "the held item is junk-class", and excludes only shops. Because `CheckQueue`'s foreign branch is a fork *inside* `if (randoSaveCheck.eligible)`, a host that never gets `.eligible` set strands a pinned progression item forever, and a paired world becomes unwinnable-by-construction with no error. Two concrete defects already slip through this predicate today, and the predicate's shape means a future check row added with `FLAG_NONE` and no hook would strand silently.

## Known

The predicate is exactly four conditions, inline in the candidate loop:

- `games/mm/2s2h/Rando/Foreign.cpp:118-132` — skip `RC_UNKNOWN`; skip `RCTYPE_SHOP`/`RCTYPE_TINGLE_SHOP`; require `RANDO_SAVE_CHECKS[id].shuffled`; require `Rando::StaticData::Items[heldItem].randoItemType == RITYPE_JUNK`. There is no test on `flagType`, on `.skipped`, or on any give-path property.
- `games/mm/2s2h/Rando/MiscBehavior/CheckQueue.cpp:40-52` — the foreign branch is nested inside `if (randoSaveCheck.eligible)`. Nothing else reaches a foreign placement. No `.eligible`, no give, ever.
- `games/mm/2s2h/Rando/StaticData/StaticData.h:20-28` — `RandoStaticCheck` is `{randoCheckId, name, randoCheckType, sceneId, flagType, flag, randoItemId}`. There is **no** per-check "has a give path" datum anywhere in the check table, so the predicate has nothing sound to consult even in principle.
- `PlaceForeignItems` is called once, from the paired branch of file creation, after the logic apply and before the spoiler write — `games/mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp:238-246`.

[...]

PR #581, opened 2026-07-31
Flip the paired MM world to Glitchless behind a deterministic attempt ladder (#500 / ADR 0010 increment 1)

spencerduncan:
# ADR 0010 increment 1, generation side (#500)

Lane: paired-MM Glitchless default + deterministic attempt ladder + loud exhaustion + the O1 option retirement. Two commits: the implementation, then an adversarial review pass that fixed one determinism defect and re-pinned the lock that was supposed to prove the ladder deterministic.

Rebased onto `main` (`0af9ccf4`), which now carries #579's ADR 0004 supersession note — the branch's own ADR 0004 edit was dropped in favour of it, keeping only an update to its "implementation in flight" parenthetical.

---

## 1. The flip (increment 1.1)

The paired MM world's `RO_LOGIC` default moves Nearly No Logic → **Glitchless** at the single resolution site (`Rando::Foreign::ResolveProfileValues`, `games/mm/2s2h/Rando/Foreign.cpp`). Existence-probe semantics are unchanged: an explicit choice of *any* mode, no-logic modes included, still wins — choosing away the proof is legitimate and is recorded in the frozen identity like everything else. The pin stays explicit rather than falling through to the StaticData default, so upstream default drift cannot silently change the paired world's rules.

Per accepted answer O11 this mirrors Ship of Harkinian's own posture (`RSK_LOGIC_RULES` → `RO_LOGIC_GLITCHLESS`).

## 2. The deterministic attempt ladder (increment 1.2)

`OnFileCreate`'s paired branch runs generation under a bounded ladder. One attempt is factored into a callable; the paired path drives it, the solo path keeps its single-shot upstream behaviour, and the RNG stream is not moved for either.

Recipe (documented at `MixPairedFinalSeedForAttempt` in `Foreign.h`, per accepted O3):

```
finalSeed(0) = Ship_Hash(decimal(masterSeed) ++ optionsString)
finalSeed(n) = Ship_Hash(decimal(masterSeed) ++ optionsString
                         ++ ":glitchless-attempt-" ++ decimal(n))   n >= 1
```

Attempt 0 is **byte-identical to the shipped pre-ladder derivation**, so a world that converged first-try before the ladder existed re-derives unchanged. The counter is an *input* to the hash, never state beside it. Bound: `kPairedGenMaxAttempts = 10`. Each retry restores the full pre-attempt save image and clears the aborted attempt's foreign placements. The winning index is recorded in the spoiler (`rsbsPairedAttempt`) and the combo record (`gComboCtx.mmPairedAttempt`, +1-displaced so 0 keeps its "unset" meaning; 4-byte carve at pinned offset 876, record size and every other offset unchanged, static-assert chain extended).

Exhaustion surfaces through the pairing-refusal path: `MM_Rando_PairOnCrossGameArrival`'s failure branch latches the slot with a new `RSBS_REFUSE_GENERATION` (session refusal, no quarantine, mirroring `RSBS_REFUSE_IDENTITY`) plus the shared-overlay toast. **Never a silent vanilla Termina.**

### 2a. Only *deterministic* dead-ends are rungs — the review's main finding

`ApplyGlitchlessLogicToSaveContext` has three failure paths, and one of them — `"Logic Generation Timeout"` — is a **10s wall-clock abort**: a function of machine speed and load, not of the seed.

As first implemented the ladder caught it like any other dead-end and rolled to the next rung. That means the same seed and settings would converge on attempt 0 on a fast machine and on some later attempt on a slow one — **two players, one frozen identity, two different worlds**. Note this is a hazard the ladder *created*: before it, a timeout produced a vanilla revert and no world at all.

Fixed: the timeout throws `Rando::Logic::GenerationTimeout` (derived from `std::runtime_error`, so every existing `catch (const std::exception&)` behaves exactly as before — this adds a discriminator, not a new escape path) and the ladder stops cold on it, taking the loud path instead of silently delivering a different world. Every other failure the fill raises is a deterministic function of the seeded RNG stream and remains a legitimate rung.

## 3. The O1 retirement

`RO_ACCESS_MAJORA_REMAINS` is retired exactly as accepted: the enumerator **and** the Options map row stay as a tombstone, and the pane keeps drawing the row disabled-with-reason naming the ruling. Deleting the enumerator would renumber the 44 after it and `RANDO_SAVE_OPTIONS` is indexed by that number in every already-written MM rando save; removing only the map row would shift every paired world's seed derivation (`MMOptionsString` hashes the table in row order) **and** every profile identity digest (`ProfileIdentityString` iterates the same rows), refusing shipped pairs at arrival over a tidy-up. Indices, `MMOptionsString` and `ProfileIdentityString` are all byte-stable. Rationale recorded at the row itself.

## 4. Locks

**(a) `MMPairedProfile` (redship, extended).** The paired default resolves Glitchless (Fail 56); an explicit Nearly No Logic choice survives the pin (Fail 60); a post-creation logic edit diverges and refuses without self-healing (Fail 66/67). `mm-rando-options` additionally pins the retirement (Fail 69–71: DORMANT + a reason naming "Retired").

**(b) `MMPairedAttemptGen` + `MMPairedAttemptDeterminism` (rando).** A paired generation whose first ladder attempt fails deterministically climbs one rung, converges through the real `OnSaveInit` chain, and reaches a world that is a pure function of the frozen identity. `MMPairedAttemptDeterminism` (`CMake/CheckPairedAttemptDeterminism.cmake`) runs the dispatch twice in two fresh processes and byte-diffs final seed + winning attempt + placement hash + every foreign placement. `SeedDeterminism`'s MM digest also folds `mmPairedAttempt`, so both processes must reach the same world *by the same derivation*.

> **The failing rung is injected, and that is deliberate — the original pin was a race.**
> The lock originally pinned master seed 17 as "dead-ends attempt 0, converges attempt 1". Seed 17 does not dead-end deterministically: it **times out**. Scanning master seeds 1..66 under the dead-end-prone Glitchless profile (ocarina buttons + swim + boss remains + owl statues + clock shuffle) produced 14 first-attempt failures and **every one was the wall-clock abort — not a single deterministic dead-end**. A seed pinned on a wall-clock failure is a race: a faster or less loaded machine converges it first try and the row flips to FAIL(4) on timing alone. And with §2a landed, a wall-clock seed would additionally be pinning behaviour the ladder no longer has.
>
> So one rung is injected deterministically (`Rando::Foreign::ForceShortForeignPlacements` — one short foreign-placement pass, which is the genuine #488 throw) on the all-default profile. The *cause* of attempt 0's failure is not what the lock is about; the ladder's **response** is, and that is what it now proves: pre-attempt save restored, seed re-derived by the documented recipe, a real fill and a real placement pass on the winning attempt (the injection count is asserted drained, so it cannot pass on a world that was never placed), the winning index recorded, byte-identical across two processes. The scan mode keeps the heavy profile and now documents the wall-clock discriminator (`attempts=1 converged=0` is a timeout and must never be pinned).

**(c) `MMPairedExhaustion` (rando), three legs.** An all-checks-excluded profile exhausts the 10-attempt bound at the real switch-entry arrival and refuses loudly — vanilla revert, slot `REFUSED(generation)`, unwritable, no leaked placements or provenance. A counter-leg with the fixture undone generates under the all-default (Glitchless) profile and latches nothing, proving the refusal is failure-driven. A third leg locks §2a: with the fill's budget squeezed to 1ms (a test-only override, zero on every shipping path), the arrival must stop at attempt 1 and refuse rather than climb.

### Counterfactuals — measured, not asserted

| Counterfactual | Result |
|---|---|
| Revert the Glitchless pin to Nearly No Logic | `MMPairedProfile` red at exactly **FAIL(56)** (`got 2`) |
| Drop the `RsbsSave_RefuseSlotGeneration` call | `MMPairedExhaustion` red at exactly **FAIL(6)** |
| Make a wall-clock abort an ordinary ladder rung again | `MMPairedExhaustion` red at exactly **FAIL(15)**, `attempts=10 exhausted=1` |

Each was applied, rebuilt, measured, then reverted and both tiers re-greened. The ladder-determinism counterfactual is documented at the test rather than run, because faking nondeterminism faithfully requires the bug the lock catches.

## 5. Verification

Local Windows, full ROM/archive-staged worktree, Release/Ninja. Baseline `main` built green first, then the branch.

```
ctest -L "^redship$"   73/73 passed   (incl. TestRegistrationComplete for the new rows)
ctest -L "^rando$"     18/18 passed
```

`MMPairedAttemptGen` 0.52s, `MMPairedAttemptDeterminism` 1.00s, `MMPairedExhaustion` 0.51s; `MMRandoGen`, `MMPairSwitchEntry`, `RandoDeterminism`, `SeedDeterminism` all green. Direct-run evidence: pinned seed 1 with one injected rung → `winningAttempt=1`, `mmPairedAttempt=2`, `mmFinalSeed=3FD46A8A`, `placementHash=DA33504E`, 4 foreign placements; exhaustion leg → 10/10 attempts, `REFUSED(generation)`; wall-clock leg → stopped at attempt 1, refused. Neither determinism row carries a golden digest (both are two-process byte-diffs), so nothing needed re-pinning.

clang-format 14 clean on the allowlist-enforced touched files. Generated version stamps excluded from both commits; no submodule pointers touched.

## 6. Compatibility and notes

- **A pre-flip default-profile pair whose MM half never generated will REFUSE at its next arrival**: its creation stamp froze the NNL resolution, and arrival now resolves Glitchless. Per the one-game ruling this is the intended loud behaviour, not a regression. Pairs whose MM half already generated are untouched (the frozen-save restore path never re-resolves).
- Default-profile paired worlds re-derive under Glitchless (a deliberate ADR re-pin). Explicit-choice profiles and solo MM generation are byte-identical. OoT is untouched.
- One pre-existing fixture needed correcting: `MM_Rando_HeadlessForeignDigest` pinned `RO_LOGIC=NNL` *after* the OoT half's creation event had stamped the profile digest. Under the flipped default #570's freeze machinery correctly refused that mid-dispatch flip. The fixture now writes no option-table CVars between stamp and MM generation and rides the shipped default profile.
- **Empirical finding worth carrying forward:** the Glitchless fill under the all-default profile converged attempt 0 on every seed tried, and under the heavy profile every observed failure was the 10s wall-clock abort rather than a dead-end. The ADR flags that abort as "unmeasured under creation flow and must be measured, not assumed, before increment 2 ships" (Consequences → creation-time compute budget) — this lane's measurements say it is not merely unmeasured but *dominant*, and it is the reason the ladder must discriminate. Increment 2 should treat the abort's budget as a product decision, not an implementation detail.
- `reserved[]` headroom is now **124 bytes**. ADR 0009's outstanding claims 2+4 (4+64) would leave 56 — under the 64-byte floor — so the budget table needs a row before those land (noted at the carve in `context.h`).
- The exhaustion toast is emitted but not asserted (no notification test hook); the latch and slot state are the load-bearing assertions.

Not merged; no issues closed.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8824986421.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8824824914.zip)


spencerduncan:
Counterfactuals, measured on this branch (2026-07-31, local Windows build), not asserted:

**CF1 — the Glitchless default flip lock.** One-line revert of the pin (`values[RO_LOGIC] = RO_LOGIC_NEARLY_NO_LOGIC` at the resolution site, `games/mm/2s2h/Rando/Foreign.cpp`), rebuild, `redship --test mm-paired-profile`:

```
[MM-RANDO-OPTIONS] FAIL(56): with no explicit choice, the paired profile did not default RO_LOGIC to Glitchless (ADR 0010 increment 1.1; got 2)
EXIT=1
```

**CF2 — the loud-exhaustion lock.** Dropped only the `RsbsSave_RefuseSlotGeneration(slot)` call from the arrival gate's failure branch (`games/mm/2s2h/GameExports_SingleExe.cpp`), rebuild, `redship --test mm-paired-exhaustion`:

```
[MM-EXHAUST] FAIL(6): the exhausted arrival did not surface REFUSED on the active slot (state=0) — the silent-vanilla-revert class, back again (#564 V7)
EXIT=1
```

Both edits reverted; the committed tree rebuilt and both tiers re-ran green afterward (73/73 redship, 18/18 rando). The ladder-determinism counterfactual (make the derivation consume runtime state → the two-process `MMPairedAttemptDeterminism` diff fails) is documented at the test rather than measured — faking nondeterminism faithfully requires exactly the bug the lock exists to catch.

One tier finding worth naming: the first rando-tier run on this branch went red on `RandoDeterminism`/`SeedDeterminism` — and the failure was the freeze machinery doing its job. The MM digest fixture used to pin `RO_LOGIC = Nearly No Logic` *after* the OoT half had run the creation event (which stamps `gComboCtx.mmProfileDigest` from the live CVars, #570). Under the flipped default that mid-dispatch flip became a real post-creation divergence and `ResolvePairedProfile` refused it, before the ladder (identity divergence is deliberately not retried). The fixture now writes no option-table CVars between the stamp and the MM generation, so the digest run rides the shipped default profile end to end.

Issue #596, opened 2026-08-02
ArchiveManager uses operator[] on lookup, permanently poisoning HasFile after any miss


spencerduncan:
Found during the #577 re-examination (2026-07-31): `ArchiveManager.cpp:51,68` use `operator[]` on a lookup path. `operator[]` default-inserts on a miss, so any failed lookup permanently poisons the map — `HasFile` then reports a bogus entry forever after for that key.

Same class as the resource-cache races fixed in #566's review. Should be `find()`/`contains()`.

See #577 comment issuecomment-5155292992 for the anchors.

spencerduncan:
Fork branch and PR prepared, not merged and not pinned: **https://github.com/spencerduncan/libultraship/pull/5** (branch `claude/archive-manager-lookup-insert`, stacked on `claude/gl-texture-name-overflow` = `477633a`, the commit redship currently pins).

Both sites confirmed at that commit: `ArchiveManager::LoadFile(uint64_t)` at `:51` and `GetArchiveFromFile` at `:68` reach `mFileToArchive` through `operator[]`, which default-inserts a null `shared_ptr<Archive>` on a miss. `HasFile(uint64_t)` answers from `mFileToArchive.count(hash)`, so a single failed lookup makes `HasFile()` report **true** for that path permanently, and a probe-then-load caller takes the "present" branch and gets a null `File`. Both are now `find()`.

Those two were the only remaining `operator[]` reads: `HashToString`/`HashToCString` already used `find()`, and the other three `mFileToArchive` sites (`:145` `clear()`, `:161`, `:281`) are genuine writes during archive registration.

The redship-side half of this group — #595 and #593 — is #602, verified against the **unmodified** pinned libultraship. The submodule bump for this fix is the shepherd's call.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #577, opened 2026-07-31
[Epic] Cross-game asset rendering: blocked-as-currently-built, not impossible — curated archive spike


spencerduncan:
**This epic is not scheduled work.** It exists because **ADR 0010** told it to exist: `Non-goals → future epics` item 3 (`docs/adr/0010-cross-game-logic-and-beatability.md:681-686` at origin/main `a6aa26d3`) names cross-game asset rendering **permanently blocked** and then says, verbatim, "**The epic exists to hold the 'unless the namespace is reworked' line so nobody rediscovers the blocker mid-increment; it is not scheduled work.**" It is filed on that instruction, as the standing destination for every future "can't we just draw the other game's model" proposal.

## The line this epic holds

> **One game cannot render the other game's asset, and this is a namespace fact, not a missing feature. It reopens only if the resource namespace is reworked — which is not planned.**

## Why it is blocked (verified, standing)

- **151 colliding object names** between the two archives — 382 OoT objects vs 464 MM objects, 39.5% of OoT's namespace, plus 14 overlay collisions, for 165 total (`docs/resource-namespace-audit.md:7-13`, full list at `:137`).
- **Lazy archive mounting**: the inactive game's archive is not mounted while the other game runs, so even a non-colliding name has nothing to resolve against at draw time.
- Together these mean a cross-game draw is not "a model loader away" — it needs a namespace rework across both archives and the mounting model, which is the "unless" clause above.

## The shipped answer instead

**Native, model-less presentation in both directions**: each game presents the foreign item using its own native surfaces and the item's real name, never the foreign model. That is the design that replaced the #392 Lane C `RI_RUPEE_HUGE` compromise (see #494); it is the answer, not a placeholder for this epic.

## Scope of this epic (all of it)

1. Hold the line above, and be the issue any "render the other game's item" proposal is closed as a duplicate of.
2. If a namespace rework is ever genuinely on the table for other reasons, this is where the asset-rendering consequence gets re-evaluated — and only then.
3. Keep `docs/resource-namespace-audit.md` the single standing explanation. If an in-tree comment ever claims cross-game rendering is possible or merely unimplemented, that comment is stale and correcting it belongs here.

## Dependencies

**None, and deliberately so.** Gated by no increment; blocked on a namespace rework that is not scheduled. Nothing in ADR 0010's increments 1-3 depends on it, and nothing in it depends on them.

## References

- ADR 0010 §`Non-goals → future epics` item 3 (`:681-686`)
[...]

Issue #154, opened 2026-01-27
Resource archive hot-swapping for cross-game switches


spencerduncan:
When switching from OoT to MM, MM needs its own resource archives (mm.o2r) loaded into the ResourceManager. Currently OoT loads oot.o2r and soh.o2r during InitOTR().

**Needed:**
- Load MM archives when MM starts (via `ArchiveManager::AddArchive`)
- Verify `AddArchive` works post-init (single-exe shares the ResourceManager)
- Consider unloading OoT-specific archives to free memory, or keeping both loaded

**Context:** Part of the cross-game switching lifecycle (#145). The GameRunner (#153) now properly suspends OoT instead of shutting it down, keeping the ResourceManager alive.

PR #602, opened 2026-08-02
Stop curated-archive collisions and switch-clobbered mods (#595, #593)

### LINKED ISSUE #593: OoT-side mod archives are silently clobbered on every game switch
Found during the #577 re-examination (2026-07-31): on the OoT side, mod archives are silently clobbered on every game switch. A player's mods stop applying (or apply inconsistently) after the first crossing, with no diagnostic.

Verify at the archive-mounting path in `rsbs/src/main.cpp` / `EnsureGameArchivesLoaded` and the OoT-side mount sequence; the #577 comment (issuecomment-5155292992) has the anchors.

Severity: user-visible for anyone using mods, silent.

### LINKED ISSUE #595: soh.o2r and 2ship.o2r collide on 595 paths (21 differing), so some textures depend on game-switch order
Found during the #577 re-examination (2026-07-31): `soh.o2r` and `2ship.o2r` — the two curated custom-asset archives — collide with each other on **595 paths, 21 of which differ in content**. Because both are mounted into one flat ArchiveManager, which copy wins depends on mount order, which depends on which game booted first and whether a switch has happened.

Observable consequence identified: chest-corner textures depend on switch order. Anything else among the 21 differing paths has the same exposure.

Fix shape is presumably to namespace the two curated archives (we generate both, so the paths are our choice) or to make the lookup archive-scoped (`ResourceMgr_ListFilesForGame` / `Combo_ArchivePathIsMM` already do this for audio). See #577 comment issuecomment-5155292992.

spencerduncan:
Closes #595. Closes #593.

Two archive-layer defects found by the #577 re-examination, both downstream of the same property: `ArchiveManager::AddArchive` overwrites `mFileToArchive[hash]` unconditionally, so resolution is **last-added-wins, globally, with no priority field** (`libultraship/src/ship/resource/archive/ArchiveManager.cpp`). In a single-executable build both games' archives live in one flat manager, so "who wins" is a function of mount order — and mount order is a function of which game booted first and whether a switch has happened.

The third issue in this group, #596, is fork-side: **spencerduncan/libultraship#5**, stacked on the commit this repo pins. The submodule pointer is deliberately **not** bumped here.

---

## #595 — the two curated archives collided on 595 paths, 21 differing

`soh.o2r` and `2ship.o2r` are the archives we generate ourselves from in-tree custom assets (`--norom --custom-assets-path`). Re-derived at HEAD by enumerating both as zips: **1280 and 840 paths, 595 shared, 21 differing in content.** All 21, and what each is:

| # | Path | What it is |
|---|---|---|
| 1–3 | `accessibility/texts/{filechoose,kaleidoscope,scenes}_eng.json` | OoT's screen-reader text banks, loaded at `games/oot/soh/Enhancements/tts/tts.cpp:1156` as `accessibility/texts/<bank><lang>`. 2Ship ships stubs of the same three — its `scenes_eng.json` is literally `{"0": ""}`, 16 bytes against OoT's 3940 — and **no MM code reads them**: 2Ship has no TTS at all. When 2ship.o2r won, OoT's scene / file-select / pause announcements went silent. |
| 4–10 | `objects/object_box/gBoxChestCorner{Health,Lesser,Major,Mask,SkullToken,SmallKey,StrayFairy}Tex` | Chest-appearance-matches-contents corner textures. Genuinely different art per port (2140 B vs 2128 B). OoT reaches them from inside soh.o2r's own display lists (`mat_gChestBody*DL_Sides`, `SetTextureImage Path=`); MM reaches them from C, `gSPSegment(0x09, …)` at `games/mm/2s2h/Rando/ActorBehavior/EnBox.cpp:97-126`. This is the pair the issue observed. |
| 11–17 | `objects/object_box/gBoxChestLock{…}Tex` | Same fourteen-way split, the lock plate (4188 B vs 4176 B). |
| 18–21 | `shaders/{directx/default.shader.hlsl, metal/default.shader.metal, opengl/default.shader.fs, opengl/default.shader.vs}` | Fast3D's Prism shader sources, fetched by **fixed** archive path by the render backends (`gfx_opengl.cpp:312,370`, `gfx_direct3d11.cpp:1331`, `gfx_metal_shader.cpp:256`). OoT's copy is copied out of `libultraship/src/fast/shaders/` at archive-generation time; MM's was a committed snapshot. **These four are a platform artifact today, and it is worth being exact about it:** all four blobs are byte-identical in *content* between MM's tracked copy and libultraship's at the pinned commit, so on Linux the real content-divergence count is **17**. On Windows the byte difference is exactly the line count (6687−6477 = 210 lines, same identity for all four) because this repo's `.gitattributes` forces `eol=lf` while the submodule has none and `core.autocrlf=true` — so soh.o2r packed CRLF and 2ship.o2r packed LF. The *snapshot* is the actual hazard: nothing kept the two in sync, and a libultraship bump would silently diverge them for real, in a process with **one** renderer for both games. |

### Fix — three different answers, because the 21 are three different situations

The lookups cannot be made archive-scoped from this side: `ResourceManager` already keys its cache on `{Path, Owner, Parent}` but drops `Parent` at `ResourceManager.cpp:141` (`LoadFileProcess(identifier.Path)`), which is a fork change. So the divergence is removed at the source instead, which is legitimate here because **we generate both archives, so the paths are our choice**.

1. **Shaders — make them the same bytes, structurally, and stop keeping a second copy.** `games/mm/assets/custom/shaders/` is now untracked and gitignored, exactly like `games/oot/assets/custom/shaders/`, and `ExtractMMAssets` / `Generate2ShipOtr` do `rm -r -f` + `copy_directory` from `libultraship/src/fast/shaders/` before packing — the same two steps the two OoT targets already run. Both curated archives then get their shaders from the one libultraship source, on every platform, so mount order cannot pick a different shader and a libultraship bump cannot leave MM on a stale renderer.

   The untracking is load-bearing, not tidiness. An earlier revision of this PR kept MM's four copies tracked and only added the copy step; **that dirties the working tree on every Windows build** — `.gitattributes eol=lf` does not reach into the submodule, so with `core.autocrlf=true` the LUS sources are CRLF on disk and the tracked LF destinations come back as four ` M ` entries in `git status`. Reproduced directly (copy LUS's shaders over the tracked files → four modified files), and verified fixed: after a full `ExtractAssets` + `ExtractMMAssets` + `Generate2ShipOtr` run on this branch, `git status` shows only the usual generated version stamps.

2. **Accessibility texts — delete the dead stubs.** MM has no TTS and no consumer for these; they existed only to shadow OoT's real banks. Verified by grep across the whole tree: the single reader anywhere is `games/oot/soh/Enhancements/tts/tts.cpp:1156`, and OoT ships its own `accessibility/texts/*` under `games/oot/assets/custom/`, so removing MM's copies leaves exactly one source for each bank.

3. **Chest textures — game-scope the MM copies.** Both are live, both are wanted, so MM's fourteen move to a game-scoped subtree: `games/mm/assets/custom/mm/objects/object_box/` → archive path `mm/objects/object_box/…`, with the fourteen `__OTR__` literals in `games/mm/assets/2s2h_assets.h` repointed. The C symbol names are unchanged, so `EnBox.cpp` is untouched. `mm/` mirrors the game tags `ResourceMgr_ListFilesForGame` / `Combo_ArchivePathIsMM` already use in production for audio, and the header states the rule: anything else 2ship.o2r would ship under a colliding path with differing content belongs there too.

Deliberately **not** moved: MM's vanilla `objects/object_box/` (that is mm.o2r, untouched) and the eight non-colliding custom textures in the same folder.

---

## #593 — mod archives were silently revoked on every switch

A user mod overrides a base asset *only* by being mounted after it; both ports mount `mods/` last for that reason. `EnsureGameArchivesLoaded` re-adds the destination game's base archives on every cross-game switch — it has to, that is the #154 fix — which put them back on top. OoT feels it on the first return trip (it mounts mods late, at GUI init, `mod_menu.cpp` `UpdateModFiles`); MM on its second arrival (it mounts them in the initial list, `BenPort.cpp`). Either way the failure is silent: the mods stay listed as enabled and simply stop applying.

Fix: a small per-game registry (`src/common/mod_archives.{h,cpp}`) that each port feeds at mount time, in mount order, and which the switch path replays **after** the base re-add. Re-globbing `mods/` there instead would have been wrong — OoT's enabled set is a CVar-persisted *subset* of the folder, so a glob would resurrect archives the player disabled, and mod order is user-controlled (the mod menu reorders it deliberately), which is why registration is idempotent *without* reordering. Only the target game's mods are replayed: the base archives just re-added are the target's, so the target's mods are exactly what got clobbered.

Registration cannot go stale within a session: OoT's `EnableMod`/`DisableMod` do not mount or unmount at runtime (both `AddArchive`/`RemoveArchive` calls there are commented-out TODOs upstream), so the registry is exactly the set the port actually mounted.

Any mod that cannot be re-applied now says so on stderr, naming the archive and stating that it will not apply. Silence was the whole defect.

`EnsureGameArchivesLoaded` is renamed `Combo_EnsureGameArchivesLoaded` and declared in `mod_archives.h` — not for reuse, but so the lock below drives the *production* function instead of a copy of it.

---

## Locks

Both rows are archive-layer, display-free and ROM-free (they need only the two `--norom` curated archives, which CI already stages into this tier for #562), and both SKIP rather than fail when those archives are absent — the same policy as `ZipContention`, because the netplay-relay job re-runs this label archive-less on purpose.

**`CuratedArchiveOrder` (#595)** mounts soh.o2r and 2ship.o2r into two independent `ArchiveManager`s, **soh-first and 2ship-first**, and requires every path either archive contributes to resolve to identical bytes both ways. It states the invariant rather than enumerating the 21, so it needs no allowlist and catches a twenty-second colliding path the day someone adds one. Two anti-vacuity guards, because "no differences found" is the pass condition: each archive must expose a plausible path count, and the two must actually *share* paths — without the second, the row would go green if a future change simply stopped generating one of the archives.

The row reads through `GetArchiveFromFile` + `Archive::LoadFile(const std::string&)` rather than `ArchiveManager::LoadFile`, because the latter resolves the archive and then calls `Archive::LoadFile(hash)`, whose first act dereferences the **global** `Context::GetInstance()->GetResourceManager()->GetArchiveManager()` (`O2rArchive.cpp:16-19`) — a null deref off the singleton. Confirmed safe for this row specifically: `Archive::Load()` only reaches that global when the archive carries a `version` file, and the exporter writes `version` **only** into ROM-extracted archives (`OTRExporter/OTRExporter/Main.cpp:169` vs `:214`, where the custom archive gets `portVersion` alone). Verified against the generated pair — neither curated archive contains `version`.

**`ModArchiveSurvivesSwitch` (#593)** drives `Combo_EnsureGameArchivesLoaded` against the real shared `ArchiveManager`, with a copy of a staged archive standing in for a mod, and asserts the mod owns the contested path afterwards. It starts from the *strictest* state — the base archive explicitly on top — so a switch path that did nothing at all would fail. Its own negative control then runs the identical sequence with an **empty registry** and asserts the base wins there, so the positive leg is provably detecting the replay and not an incidental ordering. It also covers the registry contract: duplicates do not reorder, null / empty / `GAME_NONE` are rejected, per-game lists do not leak, out-of-range returns NULL.

The row snapshots the shared manager's archive list up front and restores it with `SetArchives` on the way out, rather than just removing the stand-in mod. It mounts the base archive twice and everything `Combo_EnsureGameArchivesLoaded` adds twice, and `AllTests` runs every row in one process — leaving `oot.o2r`/`soh.o2r` re-added on top would hand a later MM row OoT's copy of the 151 paths `oot.o2r` and `mm.o2r` both carry.

### Counterfactuals actually run

**#595, re-run on the final state of this branch.** Same binary, same row, with only the chest-texture scoping reverted (`mm/objects/object_box/*` moved back to `objects/object_box/*`) and `2ship.o2r` regenerated from that tree:

```
[curated-archive-order] soh.o2r: 1280 paths, 2ship.o2r: 836 paths
[curated-archive-order] shared paths: 591
[curated-archive-order] ORDER-DEPENDENT: objects/object_box/gBoxChestCornerHealthTex (soh-first 2128 bytes, 2ship-first 2140 bytes)
   … 6 more Corner entries, all 2128 vs 2140 …
[curated-archive-order] ORDER-DEPENDENT: objects/object_box/gBoxChestLockHealthTex (soh-first 4176 bytes, 2ship-first 4188 bytes)
   … 6 more Lock entries, all 4176 vs 4188 …
[curated-archive-order] FAIL: 14 of 1525 paths resolve differently depending on which curated archive was
mounted first (#595)
[TEST] FAIL: curated archive order rc=1
```

Exactly the pair the issue observed, with both sizes named, and the shared count moving 577 → 591 as the fourteen come back. Restored and regenerated:

```
[curated-archive-order] soh.o2r: 1280 paths, 2ship.o2r: 836 paths
[curated-archive-order] shared paths: 577
[curated-archive-order] PASS: 1539 paths (577 shared) resolve identically in both mount orders
```

Cross-checked outside the binary, straight off the zips: `soh 1280, 2ship 836, shared 577, differing 0`, with the fourteen chest textures present as `mm/objects/object_box/…`.

**#593** — the row's empty-registry control *is* the counterfactual and runs on every invocation:

```
[mod-survives-switch] with 1 registered mod, portVersion is owned by: …/build-cmake/rsbs_test_mod_593.o2r
[mod-survives-switch] with 0 registered mods, portVersion is owned by: …/build-cmake/soh.o2r
```

---

## Verification

Full local build (Ninja / Release, `-j4`, low priority) in a clean worktree at this branch merged with `origin/main`, with all four archives staged (both curated ones regenerated from this tree, plus ROM-extracted `oot.o2r` / `mm.o2r`), then both ctest tiers:

| Tier | Result |
|---|---|
| `-L "^redship$"` | **75/75 passed** (73 before this branch; `CuratedArchiveOrder` and `ModArchiveSurvivesSwitch` are the two new rows) |
| `-L "^rando$"` | **18/18 passed**, including `RandoDeterminism`, `SeedDeterminism` and `MMPairedAttemptDeterminism` — nothing here moves an RNG stream |

Also run locally: `check-odr-declaration-collisions.py` (64 audited duplicates, 0 new, 0 stale) and the repo-invariant suite (`159 passed, 17 skipped`). No file in `.github/clang-format-paths.txt` is touched by this branch. Submodule pointers verified unchanged at `477633a` / `a956780` / `270324d`; generated stamps (`games/{oot,mm}/src/boot/build.c`, `games/oot/properties.h`, `games/mm/windows/properties.h`) reverted before every commit.

### Known limits, stated plainly

- The switch path repairs the **archive layer**. `ResourceManager`'s resource cache is not invalidated on a switch (nothing in the switch path calls into it), so an asset already loaded and cached before a crossing keeps whatever bytes it had. That is true with or without this change; the fix makes every *fresh* resolution after a switch correct again.
- The shader leg of #595 is a no-op on Linux, where the tracked copies were already byte-identical to the submodule's. CI's `generate-rsbs-otr` builds both curated archives on Linux, so the two new rows exercise the chest-texture and accessibility legs there; the EOL divergence and its fix are only observable on a Windows build, which is where the counterfactual above was run.
- Two stale in-tree comments still misdescribe the mounting model and are left for #577: `games/oot/soh/Enhancements/Restorations/GraveHoleJumps.cpp:91-92` ("OoT archives are unloaded when switching to MM" — nothing ever unloads them; the conclusion it draws is still safe) and `games/mm/2s2h/Rando/MiscBehavior/CheckQueue.cpp:70-75`.
- Pre-existing and not touched here: `AddArchive(path)` constructs a fresh `Archive` every call and never dedupes, so each crossing leaks an archive object plus a `zip_t`. #593's fix adds one more `AddArchive` per switch per mod on top of that.

### Build Artifacts
  - [rsbs-windows.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8830589635.zip)
  - [rsbs-linux.zip](https://nightly.link/spencerduncan/redshipblueship/actions/artifacts/8830483572.zip)


spencerduncan:
CI green: `build-linux`, `build-windows`, `netplay-relay`, `generate-rsbs-otr`, `clang-format`, `clang-tidy`, `odr-declaration-gate`, `test`, `link-check` all pass (`build-macos` / `create-release` skip as usual).

Two of those are load-bearing for this change specifically:

- **`generate-rsbs-otr`** builds `soh.o2r` and `2ship.o2r` from this tree, so it exercises the moved chest textures and the new shader copy step in `ExtractMMAssets` / `Generate2ShipOtr`.
- **`netplay-relay`** re-runs the `redship` label **without archives** on purpose (the tier's archive-less control, #562). Its pass is the evidence that both new rows self-skip cleanly rather than going red there.

### Not in this PR — #596 needs a shepherd action

The third defect in this group is fork-side and is prepared but deliberately unmerged and unpinned here: **spencerduncan/libultraship#5**, stacked on `claude/gl-texture-name-overflow` (`477633a`), which is the commit this repo currently pins. It changes `ArchiveManager::LoadFile(uint64_t)` and `GetArchiveFromFile` from `operator[]` to `find()` — both are lookups, and `operator[]` default-inserts, so one failed lookup made `HasFile()` (which answers from `mFileToArchive.count()`) return true for that path for the rest of the process lifetime.

The submodule pointer is **not** touched by this PR. Everything above was verified against the pinned libultraship, unmodified.

Worth noting since it is adjacent: the new lock reads through `GetArchiveFromFile` + `Archive::LoadFile(const std::string&)` rather than `ArchiveManager::LoadFile`, because the latter resolves the archive and then dereferences the **global** manager's `HashToString` (`O2rArchive.cpp:16-19`) — which is a null deref for any `ArchiveManager` that is not the singleton's. That is a real sharp edge in the same file #596 touches, but it is a separate shape (a hard-wired global in an otherwise instance-scoped API) and is left alone here.

